PR #14, opened 2021-05-23
Some changes to the infoview

gebner:
I've been trying out the nvim extension, and these are some minor issues that I've run into.  Please tell me if I should rather put them in my init.vim instead.

* The LSP client used to run on the infoview because it has the lean filetype.  This is annoying because it shows error messages in the infoview, so I've changed it to a leaninfo filetype.
* The infoview opened on the left side for me, and more importantly, the infoview was the current window so I always had to switch back to the lean file.  (Maybe that's because I don't have `splitright` set.)
* It took me a while to figure out why the infoview had such a huge delay.  Apparently the infoview updates are tied to the updatetime setting, which defaults to 4000ms.  I've put in a reasonable default value to prevent surprises.  (Although that's probably not the best way since you can't override it.)
* The key feature of the infoview in vscode is that it show both the goals as well as the error messages on the current line.  This is important because most lean errors are multi-line messages, but the usual diagnostic displays focus on showing only single-line messages, making it hard to read the full error.  Furthermore, many error messages are morally goals anyhow (e.g. the "don't know how to synthesize placeholder" error shows the goal at the underscore).  Besides, there's enough screen real estate in the infoview.

### DIFF
--- a/ftplugin/lean/infoview.vim
+++ b/ftplugin/lean/infoview.vim
@@ -1,0 +1,2 @@
+" required for infoview
+setlocal updatetime=100
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -17,18 +17,27 @@ local _DEFAULT_WIN_OPTIONS = {
 
 function M.update(infoview_bufnr)
   local _update = vim.b.lean3 and lean3.update_infoview or function(set_lines)
+    local current_buffer = vim.api.nvim_get_current_buf()
+    local cursor = vim.api.nvim_win_get_cursor(0)
     local params = vim.lsp.util.make_position_params()
     -- Shift forward by 1, since in vim it's easier to reach word
     -- boundaries in normal mode.
     params.position.character = params.position.character + 1
     vim.lsp.buf_request(0, "$/lean/plainGoal", params, function(_, _, result)
-      if not (result and result.goals) then
-        return
-      end
       local lines = {}
-      for _, each in pairs(result.goals) do
-        vim.list_extend(lines, vim.split(each, '\n', true))
+
+      if result and result.goals then
+        for _, each in pairs(result.goals) do
+          vim.list_extend(lines, vim.split(each, '\n', true))
+          vim.list_extend(lines, {''})
+        end
       end
+
+      for _, diags in pairs(vim.lsp.diagnostic.get_line_diagnostics(current_buffer, cursor[0])) do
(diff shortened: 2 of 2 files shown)


Julian:
Oh! Awesome! This is great, thank you! I think all definitely are mergable.

Some quick comments anyhow:

> The LSP client used to run on the infoview because it has the lean filetype. This is annoying because it shows error messages in the infoview, so I've changed it to a leaninfo filetype.

Yeah, that's reasonable to me. You lose some syntax highlighting in the infoview window which I wanted to write at some point (i.e. syntax highlighting for the format the infoview window uses itself for terms).

> The infoview opened on the left side for me, and more importantly, the infoview was the current window so I always had to switch back to the lean file. (Maybe that's because I don't have splitright set.)

Good catch, I have `splitright` set yeah but glad this is portable now.

> It took me a while to figure out why the infoview had such a huge delay. Apparently the infoview updates are tied to the updatetime setting, which defaults to 4000ms. I've put in a reasonable default value to prevent surprises. (Although that's probably not the best way since you can't override it.)

Oy, sorry about that, I even knew this was an issue, I set it [in my own config](https://github.com/Julian/dotfiles/blob/main/.config/nvim/ftplugin/lean.vim) but hadn't gotten a minute to write the upstream config (as you say it could use a bit of configurability but can do that later), so glad you did!

> The key feature of the infoview in vscode is that it show both the goals as well as the error messages on the current line. This is important because most lean errors are multi-line messages, but the usual diagnostic displays focus on showing only single-line messages, making it hard to read the full error. Furthermore, many error messages are morally goals anyhow (e.g. the "don't know how to synthesize placeholder" error shows the goal at the underscore). Besides, there's enough screen real estate in the infoview.

I'm totally OK with this change and defer to folks like you with way more vscode and Lean experience on what's useful :) so yeah thanks -- FWIW, in my own config, I use a keybinding for [`vim.lsp.diagnostic.show_line_diagnostics`](https://github.com/Julian/dotfiles/blob/main/.config/nvim/lua/config/lsp.lua#L38) to expand the 1-line diagnostic message to the full Lean one, but definitely happy with how you've done it as well, I've hacked in and out diagnostic info in the infoview at times to see how I like it.

Thanks again! Really appreciated.

Julian:
And hopefully I've said it clearly enough but I'm obviously still very new! So if you see anything else you feel strongly should be functionally different please feel free to suggest it.

PR #15, opened 2021-05-23
treesitter: do not highlight anonymous constructors

gebner:
Everything inside brackets `⟨let x := 42; x + x⟩` was highlighted differently, which is surprising and not how the vscode and emacs highlighting works.


Julian:
Thanks!

PR #16, opened 2021-05-23
infoview: add highlighting

gebner:
infoview_highlighting

I'm not sure if picking the colors manually in the syntax file is the right way to go though.

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -15,6 +15,14 @@ local _DEFAULT_WIN_OPTIONS = {
   wrap = true,
 }
 
+local _SEVERITY = {
+  [0] = "other",
+  [1] = "error",
+  [2] = "warning",
+  [3] = "information",
+  [4] = "hint",
+}
+
 function M.update(infoview_bufnr)
   local _update = vim.b.lean3 and lean3.update_infoview or function(set_lines)
     local current_buffer = vim.api.nvim_get_current_buf()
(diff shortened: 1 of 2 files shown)


Julian:
Oo pretty...

I suspect the right way for the highlighting stuff might involve `vim.lsp.protocol.DiagnosticSeverity` and [this code in `vim.lsp.diagnostic`](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/diagnostic.lua#L925-L933) which seems like it's handling mapping those to highlight links.

Let me know if that seems right to you too, I didn't stare too closely yet. But this is awesome!

Julian:
(Also sent a push rights invitation, but of course also happy to donate this to the org if it's useful)

lua/lean/infoview.lua
```diff
@@ -27,15 +35,22 @@ function M.update(infoview_bufnr)
...
         end
       end
 
-      for _, diags in pairs(vim.lsp.diagnostic.get_line_diagnostics(current_buffer, cursor[0])) do
-        vim.list_extend(lines, vim.split(diags.message, '\n', true))
-        vim.list_extend(lines, {''})
+      for _, diag in pairs(vim.lsp.diagnostic.get_line_diagnostics(current_buffer, cursor[0])) do
+        local start = diag.range["start"]
+        local end_ = diag.range["end"]
+        vim.list_extend(lines, {'', string.format('▶ %d:%d-%d:%d: %s:',
```
Julian, line 51: I think we need a `+1` for the character too right? Otherwise I see 0 based indices if I have `#check 13` at the beginning of a line.

gebner: You're right!

gebner:
> I suspect the right way for the highlighting stuff might involve `vim.lsp.protocol.DiagnosticSeverity` and [this code in `vim.lsp.diagnostic`](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/diagnostic.lua#L925-L933) and [here](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/diagnostic.lua#L97-L112) which seems like it's handling mapping those to highlight links.

Ah, that's where those mysterious guifg-only highlighting definitions come from!  I've changed the highlighting links to the LspDiagnosticsDefault* classes.

Julian:
Nice! Merge at will I think, hooray!

gebner:
> (Also sent a push rights invitation, but of course also happy to donate this to the org if it's useful)

If you want, we can transfer the repo to the leanprover-community organization.  But that's only a name change, you'd still be the owner and on the hook for maintenance. :-)

Quite a few people have shown interest in a vim plugin, so I wouldn't be surprised if this becomes as popular as the emacs mode.  With Lean 3 the vim plugin wasn't particularly nice to use.  Somewhat complicated installation aside, I didn't know enough vimscript to implement all the bells and whistles we have in the vscode extension.  It seems much more feasible to develop a usable version now with the native LSP server in Lean 4, the native LSP client in neovim, Lua scripting, and someone on board who actually knows neovim.

Julian:
> If you want, we can transfer the repo to the leanprover-community organization. But that's only a name change, you'd still be the owner and on the hook for maintenance. :-)

Cool. Yeah I think that'd be nice then, certainly do want to keep maintaining it and don't mind doing so so far, it's what I use myself, but it would definitely benefit from others who may come along knowing more Lean than I do and seeing it in the org, as it already has from you today.

Julian:
And yeah.. writing vimscript is... ech.

Issue #27, opened 2021-05-27
Infoview should scroll up automatically


gebner:
When you scroll down the infoview, it remembers the position.  When the infoview then updates and the new content is shorter than the old one, you have an empty infoview.  In that case it scroll back up again so that you at least see that it's not empty.

PR #29, opened 2021-05-27
Reset the cursor to the top of the infoview when updating.

Julian:
Closes: #27

Bit hairy getting some testing in place, but @gebner care to review this does what you expect?

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -32,7 +32,7 @@ function M.update(infoview_bufnr)
     -- Shift forward by 1, since in vim it's easier to reach word
     -- boundaries in normal mode.
     params.position.character = params.position.character + 1
-    vim.lsp.buf_request(0, "$/lean/plainGoal", params, function(_, _, result)
+    return vim.lsp.buf_request(0, "$/lean/plainGoal", params, function(_, _, result)
       local lines = {}
 
       if result and result.goals then
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -61,6 +61,11 @@ function M.update(infoview_bufnr)
   return _update(function(lines)
     vim.api.nvim_buf_set_option(infoview_bufnr, 'modifiable', true)
     vim.api.nvim_buf_set_lines(infoview_bufnr, 0, -1, true, lines)
+    -- HACK: This shouldn't really do anything, but I think there's a neovim
+    --       display bug. See #27 and neovim/neovim#14663. Specifically,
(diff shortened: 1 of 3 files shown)


gebner:
Hmm, unfortunately this does not work as well as I'd hoped.  The cursor jumps back to the top as soon as you switch back to the window for the lean file.  Maybe it just takes getting used to, but it is very counter-intuitive that switching windows moves the cursor.  The infoview moves to the top while I'm still reading it.
```lean
-- long trace message for testing
#eval let y := (List.range 100).map fun x => dbg_trace x; x + 1 dbg_trace "this is important" y.length
```
Maybe a better heuristic is to move the cursor to the top only when you move to a new line.

The behavior of scrolloff=10 would also be okay, but that doesn't seem to work for automatic updates.

Julian:
Aha, ok, I think I follow, you want it to move just if the new contents are shorter than the window length? Or do you want it to move even if they're long but have changed? (Whereas now it moves always even if nothing whatsoever changes in the infoview contents, just CursorHold fires again and it repopulates)

I think I can imagine both of those other behaviors being the expected one in different scenarios, right?

gebner:
> Aha, ok, I think I follow, you want it to move just if there are some new infoview contents are shorter than the window length?

My main issue with the current (= main branch) behavior is that it is easy to miss errors: you scroll down a long trace message. 
 Then you move a line up, the infoview is empty, and now you think that everything is fine or that lean is still processing the file.

So for me it's only important to scroll up when the content changes and becomes shorter.

> Or do you want it to move even if they're long but have changed?

Ideally not.  When I look at the middle of a long trace message and it's updated to similarly long message, then chances are I still want to see the same part, and see if anything's changed.

> I think I can imagine both of those other behaviors being the expected one in different scenarios, right?

Good question, and I can't really tell.  That's why I'd like to keep the automatic movements to a minimum. Scrolling the infoview has never bothered me in vscode, but it's also never empty.

Julian:
OK, I think I convinced myself this is actually likely a neovim bug, which I filed as neovim/neovim#14663.

I found a way to hack around it for now, which I pushed to this PR (it actually I think has nothing to do with the cursor position but I'm not 100% and didn't go off and read the neovim C source yet).

Feel free to have another look, hopefully this is closer to what you expect, though perhaps we could be even more sticky on restoring line position. But in trying your long trace example and adding some other lines around it, I suspect this is an improvement.

gebner:
This is exactly what I wanted!  Thanks!

Issue #20, opened 2021-05-25
Closing the infoview buffer causes errors


gebner:
When I (accidentally) close the infoview buffer, the following error repeats very quickly:
```
Error executing vim.schedule lua callback: /home/gebner/lean.nvim/lua/lean/infoview.lua:61: Invalid buffer id: 3
```

It would also be nice to have commands to reopen or toggle the infoview.

Julian:
Toggling is present already -- it's mapped to [<localleader>i](https://github.com/Julian/lean.nvim/blob/5a3e9dfa59ed7d5012f13458196f87e917d13932/lua/lean/init.lua#L30) if you have `mappings = true` in the config (or if you called `.use_suggested_mappings()`).

The first one is a known issue though yeah -- I put it in a subcomment of #17 ("automatically disable the infoview when it is :q'ed") but probably better to leave this open and have it be its own issue.

Shouldn't be hard to fix.

gebner:
Then I misread the subcomment!  I thought you meant calling `:q` on the lean buffer would automatically close the infoview, which would also be useful.

Julian:
Ah! Yes agree on that one too -- I think that one we should make be configurable, I go back and forth on whether I want vim to autoquit if the infoview is the last buffer open -- sometimes I mean to continue opening some lean file and sometimes I'm done, but if we make it configurable I think that one's also a good idea!

Issue #33, opened 2021-05-30
Be able to run Lean 3 and Lean 4 buffers (with corresponding LSP clients) simultaneously


Julian:
(A prerequisite here is PRing a change to [`nvim-lspconfig`](https://github.com/neovim/nvim-lspconfig) and/or doing the appropriate dispatch here within this repo.)

Julian:
Implemented by @rish987 in #52, hooray!

Issue #13, opened 2021-05-02
Properly detect lean version for files not relative to cwd


Julian:
Right now when opening /foo/bar/baz, we still look for `leanpkg.toml` in cwd.

PR #35, opened 2021-06-02
Support for simultaneous lean3/lean4 (in separate windows/tabs)

rish987:
Depends on a merge of [nvim-lspconfig](https://github.com/rish987/nvim-lspconfig) that adds lean4 as a new language server.


README.rst
```diff
@@ -112,35 +133,22 @@ In e.g. your ``init.lua``:
...
             max_width = 79,
         },
-        -- Enable the Lean language server?
+        -- Enable the Lean3/Lean4 language servers?
         --
         -- false to disable, otherwise should be a table of options to pass to
         --  `leanls`. See https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#leanls
         -- for details though lean-language-server actually doesn't support all
         -- the options mentioned there yet.
         lsp = {
```
gebner, line 142:
```suggestion
        lsp3 = {
```
Otherwise it looks like this section applies to both versions.

README.rst
```diff
@@ -112,35 +133,22 @@ In e.g. your ``init.lua``:
...
-                buf_set_option('omnifunc', 'v:lua.vim.lsp.omnifunc')
-            end,
+            on_attach = on_attach,
             cmd = {"lean-language-server", "--stdio", '--', "-M", "4096"},
         }
+
+        -- replace the path below with a path to your Lean4 executable
+        lsp4 = {
+            on_attach = on_attach,
+            cmd = {"/path/to/lean4/executable/lean", "--server"},
```
gebner, line 150:
```suggestion
            cmd = {"lean", "--server"},
```
Can we make this the default?  `elan` should always pick the right version.  The vscode and emacs extensions also call `lean` by default.

ftplugin/lean4/switch.vim
```diff
@@ -0,0 +1,39 @@
...
+  finish
+endif
+
+let b:switch_definitions = [
+    \ g:switch_builtins.true_false,
+    \ ["#check", "#eval", "#reduce"],
+    \ ["tidy", "suggest", "hint", "linarith", "library_search"],
+    \ ["rw", "simp", "simp?"],
+    \ ["cases", "rcases", "obtain"],
+    \ ["norm_cast", "push_cast"],
```
gebner, line 14: At least these are lean3/mathlib-specfic.

indent/lean.vim
```diff
@@ -0,0 +1,6 @@
+if exists("b:did_indent")
+  finish
+endif
+let b:did_indent = 1
+
+setlocal indentkeys+==end,==begin
```
gebner, line 6: Can we make this file version-specific as well? begin/end no longer exists in Lean 4.

lua/lean/lsp.lua
```diff
@@ -16,7 +20,7 @@ function M.enable(opts)
   opts.handlers = {
     ["$/lean/plainGoal"] = M.handlers.plain_goal_handler;
   }
-  require('lspconfig').leanls.setup(opts)
+  require('lspconfig').lean4ls.setup(opts)
```
gebner, line 23: I'm completely unfamiliar with the lspconfig project setup.  Is there a reason why the server definition is in nvim-lspconfig?  IMHO it would make more sense to have it here.

indent/lean.vim
gebner: Ok, so this is Lean 3-specific because it's lean and not lean4.  Then that's right.

lua/lean/lsp.lua
rish987: That enables use of the language server without this plugin. We could include the entire default config here, but that seems redundant.

README.rst
rish987: My only concern with this is that it may be misleading for simultaneous lean3/lean4 runs in particular. In this case, it seems that `lean-language-server` will attempt to run `lean --server` as well, which (unless `elan` is called internally to change the version in that environment) will be Lean 4.

rish987: We could make this the default by setting `cmd_env.PATH` for lsp3 to include Lean 3 path and exclude Lean 4 path, but perhaps that's more trouble than it's worth.

rish987: [Allowing the executable path to be given as an option](https://github.com/leanprover/lean-client-js/blob/f636b2572baf9579f12a2b90309ca81ac86bd44d/lean-language-server/src/index.ts#L30) could enable us to transfer this "full-path" requirement to lean3. Seems like the cleanest solution if we want this to be the default.

syntax/lean.vim
```diff
@@ -0,0 +1,80 @@
+" Vim syntax file
```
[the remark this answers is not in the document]
rish987: Updated.

ftplugin/lean4/switch.vim
rish987: Gotcha, I just copied this from `ftplugin/lean/switch.vim`... I'm not personally familiar with the differences here yet, so there may be others.

README.rst
gebner: I'm happy to merge a PR to lean-language-server that adds an option for the executable path.  If you want to override the PATH, another option would be to write a short shell script wrapper (like Julian does with his "Frankenlean").

> unless elan is called internally to change the version in that environment

99% of users use elan.  A handful use Nix, and then the executable is also called `lean` and in the PATH.  The plain `lean` command is thus a very good *default* setting, it will work in most situations out of the box.

(Maybe this wasn't clear because you haven't used `elan`.  Elan includes a `lean` wrapper.  So when you call `lean` it will automatically start the right version depending on the directory you're in.  If you're in a directory which uses Lean 3, `lean-language-server` would automatically start the right version of Lean 3.)

lua/lean/lsp.lua
gebner: Ok, sounds reasonable enough but I'm still not sure why anybody would want LSP integration without syntax highlighting.

lua/lean/util.lua
```diff
@@ -0,0 +1,138 @@
+local uv = vim.loop
+local M = {}
+
+-- M.path copied from lspconfig/util to avoid creating a dependency...
```
Julian, line 4: lspconfig is Apache 2.0 licensed so in theory we're supposed to add the little blurb here.

syntax/lean.vim
```diff
@@ -0,0 +1,113 @@
...
```
Julian, line 1: Maybe it's worth considering `git subtree`'ing lean.vim rather than copy pasting it if we're going to conflict with it?

Then we at least can pull in any changes automatically

Julian (commented):
I still haven't tried actually using this yet so I don't have functional feedback yet, but just a few things that I'm thinking about:

* I'm not sure yet whether personally I expect per-window infoviews or per-tab infoviews -- what I had in mind was the latter (so switching between windows on one tab would reuse the same shared infoview). I'm guessing from scanning this that you did the former, which I guess may also be reasonable, but yeah I want to see how natural that feels
* Probably what we were discussing on Zulip affects the lspconfig PR as well -- namely when you modify the LSP there we should consider already whether to modify the `leanls` one to look for a `lean3` filetype (instead of `lean` which is what it looks for now)
* I think the default behavior you have here is the opposite of the current one if I read it right, namely the current code chooses lean 4 by default if there's no `leanpkg.toml` in the cwd, whereas I think here you default to lean 3?


Will try to leave some more comments as soon as I get a chance to set this up fully. And thanks again!

syntax/lean.vim
gebner: I'm happy to archive lean.vim and move all development here.  git subtree is really inconvenient for development though (i.e. you need to make changes in lean.nvim, then commit them to lean.vim, then update lean.nvim again).

Julian: Ok that sounds even better.

README.rst
rish987: I see - I was indeed not aware of the `lean` wrapper that `elan` introduces. That option for `lean-language-server` would be appreciated, but not at all a big deal at the moment - those of us who are too stubborn to move to an `elan`-compatible machine should be able to figure this out anyways!

rish987:
> * I'm not sure yet whether personally I expect per-window infoviews or per-tab infoviews -- what I had in mind was the latter (so switching between windows on one tab would reuse the same shared infoview). I'm guessing from scanning this that you did the former, which I guess may also be reasonable, but yeah I want to see how natural that feels

I think it makes the most sense for per-window/per-tab to be a configurable option. On larger screens, per-window infoviews could be more convenient. Working on this now...

README.rst
```diff
@@ -11,17 +11,18 @@ Prerequisites
...
-
-       Note that many simple syntactical things are not yet implemented
-       (help is of course welcome), and that ``tree-sitter-lean`` is lean
-       4-only.
+Note that many simple syntactical things are not yet implemented
+(help is of course welcome), and that ``tree-sitter-lean`` is lean
+4-only.
 
 Installation
 ------------
```
Julian, line 28: Should remove line 37 here suggesting installing lean.vim I suppose

lua/lean/init.lua
```diff
@@ -19,7 +20,10 @@ function lean.setup(opts)
   if lsp4.enable ~= false then lean.lsp.enable4(lsp4) end
 
   local infoview = opts.infoview or {}
-  if infoview.enable ~= false then require('lean.infoview').enable(infoview) end
+  if infoview.enable ~= false then
+    if opts.info_pertab then vim.g.lean_info_pertab = true else vim.g.lean_info_pertab = false end
```
Julian, line 24: Maybe just:

```suggestion
    opts.one_per_tab = opts.one_per_tab or true
```

(or the reverse default of `false` if I'm the odd one out here)

And possibly storing the mutable state in the `_infoviews` table is a bit cleaner than using a vim-global option since it's private to this module?

lua/lean/infoview.lua
```diff
@@ -131,21 +146,72 @@ function M.open()
...
+end
 
-  if M._infoviews[src_win].win then
-    vim.api.nvim_win_close(M._infoviews[src_win].win, true)
+function M.close_win(src_winnr)
+  if M._infoviews[src_winnr] then
+    vim.api.nvim_win_close(M._infoviews[src_winnr].win, true)
   end
 
   -- NOTE: it seems this isn't necessary since unlisted buffers are deleted automatically?
```
Julian, line 184: I added this I believe while seeing tests fail for different reasons and causing later tests to fail. But it's possible it indeed isn't necessary. If it doesn't seem to be, feel free to kill it.

lua/lean/infoview.lua
```diff
@@ -131,21 +146,72 @@ function M.open()
...
   --if M._infoviews[src_win].buf then
   --  vim.api.nvim_buf_delete(M._infoviews[src_win].buf, { force = true })
   --end
 
-  M._infoviews_open[src_win] = false
-  M._infoviews[src_win] = nil
+  M._infoviews_open[src_winnr] = false
+  M._infoviews[src_winnr] = nil
+  -- necessary because closing a window can cause others to resize
```
Julian, line 191: Nice catch.

lua/lean/infoview.lua
```diff
@@ -131,21 +146,72 @@ function M.open()
...
+  M.refresh_infos()
+end
+
+function M.refresh_infos()
```
Julian, line 195: Maybe make this private, i.e. 

```suggestion
local function refresh_infos()
```

unless there's really a reason someone will call it via the lua API?

lua/lean/infoview.lua
```diff
@@ -131,21 +146,72 @@ function M.open()
...
+  if not vim.g.lean_info_pertab then return end
+
+  M.close_all_info()
+
+  vim.g.lean_info_pertab = false
+
+  M.set_update()
+end
+
+function M.close_all_info()
```
Julian, line 169:
```suggestion
function M.close_all()
```

Julian (commented):
This is starting to look great!

lua/lean/infoview.lua
```diff
@@ -1,6 +1,4 @@
-local lean3 = require('lean.lean3')
-
-local M = {_infoview = nil, _opts = {}}
+local M = {_infoviews = {}, _infoviews_open = {}, _opts = {}}
```
Julian, line 1: Sorry for still working through things here, but one question -- can you elaborate on your thinking for having separate `_infoviews_open` here, which I think relates to functionality that persists infoview windows after they're semi-closed, right?

What benefit is there in supporting that extra state / how do you see it being used?

rish987: Good question - it may be wise to refactor this to instead let `_infoviews` have two fields, `data: { win: ..., buf: ...}` and `open`, and remove `_infoviews_open` - let me know if you'd like this in this PR.

If we didn't have a state indicating closure, I'm not sure how we would distinguish between an infoview simply not having been created yet (in which case `_infoviews = nil`, and we _would_ want to open it) and having been explicitly closed.

rish987: Actually, after the refactor above, we may be able to indicate this by setting `_infoviews[key].data = nil` when closed and `_infoviews[key] = nil` when erased - though doing this might be a little bit less readable.

rish987: OK I see where you might have been confused here - I moved the infoview opening code to the `M.update()` method because I thought it was necessary for this, but looking back it should be fine to move it back to `M.open()`, and add a config setting `infoview_open_default` to control whether new source windows automatically open it. In this case, I should be able to use whether or not `_infoviews[key] = nil` as an indicator of whether it is open. Let me know if you'd like this in this PR.

Julian: Right something like that is I think exactly what's confusing me yeah. I think probably it's reasonable to do on top of what's here, yeah, since I'm still trying to wrap my head around things, so I suspect doing it that way will actually simplify things a bit and make them more similar to the original diff, so worth doing I suspect. Let me first just push a simple style commit but then think go for it!

Julian:
Also, any idea what's happening in the below behavior:

[asciicast](https://asciinema.org/a/h6kXOpfX2stUsHDorHh4OUVgn)

When I set per-window infoviews, it looks like the first existing one momentarily disappears until the original file is re-entered. Do you see that same behavior?

rish987:
This is expected - when you enter per-window mode, the tab's infoview is closed, and infoviews will only be opened in other windows when you trigger `CursorHold` (i.e. `M.update()`)

lua/lean/infoview.lua
rish987: Awesome, yeah I think I've finished and yeah it does look a lot cleaner :P

rish987: Ok actually here are my updated thoughts on this - it may be better just to do the initial refactor I mentioned above, because we're eventually going to have to add persistent state to infoviews between open/close anyways when we implement e.g. pinnable/pausable messages. At that point we would be forced to have a new state field indicating open/closed. Thoughts?

lua/lean/infoview.lua
```diff
@@ -200,8 +187,8 @@ function M.update()
 end
 
 function M.enable(opts)
+  opts.one_per_tab = opts.one_per_tab or true
```
rish987, line 190: I believe this would be interpreted as a boolean rather than ternary expression (and would always evaluate to `true`).

rish987:
Alright I've done the refactor with a good amount of testing. The code is a bit longer (because of some utility functions), but I find it conceptually simpler. It's ready for you to check out here: https://github.com/rish987/lean.nvim/tree/infoview_persistence_refactor

This refactor should significantly help us in the future, as we implement many new kinds of infoview states that should persist when the window is closed/opened. Though it's up to you whether you want it to go in this PR or a future one. What we have now should be functional enough already if you want to put this off for a bit.

An unexpected immediate benefit of this refactor is that we can now preserve open/close states between `:LeanInfoPerTab` and `:LeanInfoPerWin` changes:

https://asciinema.org/a/7AfLVjFzx0F3jVS71sIXK1uJm

I'm going to do a bit more testing tomorrow morning... have to start a compile of the latest Lean nightly so I can try out the new term goal stuff

lua/lean/infoview.lua
Julian: Thanks, sloppy me, yeah should be an `if` -- will fix if it's not already on your new branch.

lua/lean/infoview.lua
Julian: I don't think I follow what you mean when you're talking about persistent state yet -- I still don't understand why we don't always just want to completely get rid of an infoview when it's closed -- for pinnable messages it seems to me all you do is temporarily/permanently unhook calling `update` on a window, and then it will stay. Maybe I'm missing something still though.

But I want to get these set of incremental changes merged from this PR before doing more functionality, unless the implementation simplifies.

rish987: This makes it more similar to VSCode, where you can close the webview without having your set of pinned messages deleted.

This will be one of the aspects of this that seems "necessary" to make this a practical tool. For example, say you're editing Lean and LaTeX side-by side. The infoview is useful for getting an idea of how to interpret the Lean code, but once you understand it you'll probably want to get rid of the infoview for more screen space as you work on your LaTeX code. Now if there's a detail that you missed, you'll want to open the infoview again to see your pinned messages without them having been deleted.

Again, the code we have now should be good for a merge, but I'm pretty certain we'll want to implement persistence in the future.

Julian: OK, let's take things step by step, if/when we get there we can make the implementation more complex, but for now I think we should stick to the simplest thing that solves the current desired functionality, which is:

* don't error when closing the window manually (#20)
* allow one-per-tab or one-per-window infoviews
* properly look for leanpkg.toml relative to the file that's open (#13)
* support lean 3 and lean 4 buffers simultaneously (#33)
* properly enforce `max_width` even when windows are closed/resized

all of which are in this PR (and which I'm trying to merge one by one essentially).

If after that we need a more complex implementation we can evolve into it.

Julian:
Making some progress here -- I pulled out the piece that fixes #20. In the process I introduced some merge conflicts, sorry, will get to fixing them as I keep merging pieces.

The next thing I want to do is remove the erase logic and alongside it `_infoviews_open` (like we discussed above) and just have closed, until we get to persistent windows.

Julian:
Aha, ok now re-watching the demo from your last comment, I see finally at least some of the functionality you were implementing beyond the above list -- you want persistent window layout even when switching back and forth too.

That makes sense to me too, and explains why things are implemented this way (with the 2 state open), but yeah a bit of this is why I'm finding understanding the implementation hard to follow step-by-step.

lua/lean/infoview.lua
```diff
@@ -46,11 +41,7 @@ end
 -- either erase infoview information from table (erase=true)
 -- or indicate it has been closed (erase=false)
 local function close_win_raw(src_idx, erase)
-  if erase then
-    M._infoviews_open[src_idx] = nil
-  else
-    M._infoviews_open[src_idx] = false
-  end
+  M._infoviews_open[src_idx] = erase and nil or false
```
rish987, line 44: It looks like this needs to be reverted as well (for the same reason mentioned earlier)

rish987: Looking into this, it seems like Lua doesn't have an actual "ternary" operator - it can just be simulated using precedence and short-circuiting in specific cases.

Julian: Thanks for fixing!

lua/lean/infoview.lua
```diff
@@ -257,13 +246,7 @@ function M.close_win_wrapper(src_winnr, src_tabnr, close_info, already_closed)
     end
   end
 
-  if close_info then
```
rish987: The removal of this distinction results in `E855: Autocommands caused commands to abort` when closing the last (Lean) window in a tab with `:close` - I don't think `WinClosed` is designed to be able to handle physically closing other windows. From `:help QuitPre`: `Can be used to close any non-essential window if the current window is the last ordinary window.` No similar support exists for `WinClosed`.

lua/lean/infoview.lua
```diff
@@ -280,8 +280,17 @@ end
 -- Close the infoview associated with the current window.
 function M.close()
   if not M.is_open() then return end
+  local infoview = M._teardown(get_idx())
```
rish987: What's being done here seems to be identical to what `close_win()` was doing previously - looks like this was an attempt to fix window closing/toggling? See latest commit.

Julian:
Jumped on some work calls which means putting this on hold at least for a few hours, but I see you're already fixing more mistakes of mine -- what I'm trying to do at a high level is two things:

* the first is in [this comment](https://github.com/Julian/lean.nvim/pull/35#discussion_r648384099), namely to incrementally understand what code is required for what functionality
* the second, really part of the first, is a lot of code (nearly all of it) was moved here into the `update` function -- that seems quite odd to me, rather than the way things were done previously where events (autocmds) essentially are added or removed when associated windows come into being. I fully expect that *something* may require this kind of thing (it may even be what you pointed out about tabs), but because I can't put my finger on what that is yet, and all the functionality I've crossed off in that comment is merged already to `main` and works with the single-infoview setup, I want to get to where it's obvious what functionality is driving a design change here. Hopefully that makes sense.

Let me review your comments (and thanks again for fixing!)

lua/lean/infoview.lua
rish987: Don't quite get it yet - feel free to have a go at it and maybe I'll understand once I see the code.

Issue #25, opened 2021-05-25
Support `$CURSOR` in abbreviations in compe and builtin expanders


Julian:
In VSCode snippets, `$CURSOR` tells VSCode where to leave the cursor when expanding the snippet (and is immediately removed from the buffer).

We support this functional behavior when [expanding snippets via snippets.nvim](https://github.com/Julian/lean.nvim/blob/5a3e9dfa59ed7d5012f13458196f87e917d13932/lua/lean/abbreviations.lua#L27-L29) but not yet in the compe.nvim or builtin expanders.

Issue #21, opened 2021-05-25
compe abbreviations do not remove the inserted slash automatically


Julian:
When using the abbreviation support in compe, writing `\mu` and completing the mu will insert the mu but leave the slash in place, resulting in `\µ` in the buffer.

I cannot tell if this is a compe bug (e.g. that it is calculating ranges to remove on word boundaries or something) or me writing the compe code we have wrongly. Defining a `.confirm()` function seems like a possible way to evade even the possible bug.

(Filing this as another "known issue", though I have not been able to figure it out... :/)

compe also isn't tested in CI, which it should be alongside the snippets.nvim tests.

PR #36, opened 2021-06-03
Add builtin abbreviation expander and customize leader character

gebner:
I didn't find the the snippets and completion plugins ergonomic enough to enter unicode characters.  It's something you do all the time, so it shouldn't require extra hotkeys.

This PR adds a simple extmarks-based expander for abbreviations (e.g. typing `\a\_0)` expands into `α₀)`).  I also made the leader character configurable (which was a widely requested feature in vscode).

There are some rough edges left (e.g. exiting insert mode while in an abbreviation will make the cursor jump), but overall I think it's good enough for everyday use and it works for me.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -93,6 +93,11 @@ In e.g. your ``init.lua``:
     require('lean').setup{
         -- Abbreviation support
         abbreviations = {,
+            -- Set one of the following to true to enable abbreviations
+            compe = false, -- nvim-compe source
+            snippets = false, -- snippets.nvim source
+            enable = false, -- built-in expander
+            -- additional abbreviations:
             extra = {
                 -- Add a \wknight abbreviation to insert ♘
                 --
@@ -104,10 +109,6 @@ In e.g. your ``init.lua``:
             -- change if you don't like the backslash
             -- (comma is a popular choice on French keyboards)
             leader = '\\',
-            -- Set one of the following to true to enable abbreviations
-            compe = false, -- nvim-compe source
-            snippets = false, -- snippets.nvim source
-            enable = false, -- built-in expander
         }
         -- Enable suggested mappings?
         --
--- a/README.rst
+++ b/README.rst
@@ -90,9 +90,7 @@ In e.g. your ``init.lua``:
 .. code-block:: lua
 
     require('lean').setup{
-        -- Enable abbreviation support?
-        --
-        -- false to disable, otherwise a table of options described below
+        -- Abbreviation support?
         abbreviations = {,
             extra = {
                 -- Add a \wknight abbreviation to insert ♘
@@ -102,6 +100,11 @@ In e.g. your ``init.lua``:
                 -- this if so desired.
                 wknight = '♘',
             },
+            -- change if you don't like the backslash
+            leader = '\\',
+            -- set true to enable
+            compe = false,
+            snippets = false,
         }
         -- Enable suggested mappings?
         --
--- a/README.rst
+++ b/README.rst
@@ -50,9 +50,10 @@ part of this plugin.
 Features
 --------
 
-* abbreviation (unicode character) insertion using either
+* abbreviation (unicode character) insertion, can also provide a
   `nvim-compe <https://github.com/hrsh7th/nvim-compe>`_ or
   `snippets.nvim <https://github.com/norcalli/snippets.nvim>`_
+  source.
 
 * Initial implementations of some editing helpers (note: no
   mappings are associated with these by default unless you call
@@ -90,7 +91,7 @@ In e.g. your ``init.lua``:
 .. code-block:: lua
 
     require('lean').setup{
-        -- Abbreviation support?
+        -- Abbreviation support
         abbreviations = {,
             extra = {
                 -- Add a \wknight abbreviation to insert ♘
@@ -101,10 +102,12 @@ In e.g. your ``init.lua``:
                 wknight = '♘',
             },
             -- change if you don't like the backslash
+            -- (comma is a popular choice on French keyboards)
             leader = '\\',
-            -- set true to enable
-            compe = false,
-            snippets = false,
+            -- Set one of the following to true to enable abbreviations
+            compe = false, -- nvim-compe source
+            snippets = false, -- snippets.nvim source
+            enable = false, -- built-in expander
         }
         -- Enable suggested mappings?
         --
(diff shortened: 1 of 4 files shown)


gebner:
I tried to look around a bit to see if there are any plugins doing this kind of unicode abbreviation.  The only one I could find was [julia-vim](https://github.com/JuliaEditorSupport/julia-vim#latex-to-unicode-substitutions), and they have also rolled their own.

Julian:
Nice! Yeah I've spoken to the Julia folks briefly. They actually were interested in the ways I implemented, but I agree there's still a bit missing, especially given stuff like #21 which is very annoying. (Other than that though just to be sure, you know the autocomplete plugins support autoinsert if you like that behavior right? So they'll autoconvert \l to the arrow without any hotkeys if you enable that).

This seems nice though regardless, will give it a shot.

gebner:
> (Other than that though just to be sure, you know the autocomplete plugins support autoinsert if you like that behavior right? So they'll autoconvert \l to the arrow without any hotkeys if you enable that).

Yes, but I have actually disabled the automatic autocompletion because I found it distracting (i.e. I have to press a shortcut to even show the popup).  Putting abbreviations in the same system makes it hard to configure them differently.

README.rst
```diff
@@ -102,6 +101,13 @@ In e.g. your ``init.lua``:
                 -- this if so desired.
                 wknight = '♘',
             },
+            -- change if you don't like the backslash
+            -- (comma is a popular choice on French keyboards)
+            leader = '\\',
+            -- Set one of the following to true to enable abbreviations
+            compe = false, -- nvim-compe source
```
Julian, line 108: Maybe reorder these to the top of the table here, since without setting them none of the rest matters much.

README.rst
```diff
@@ -90,10 +91,13 @@ In e.g. your ``init.lua``:
...
     require('lean').setup{
-        -- Enable abbreviation support?
-        --
-        -- false to disable, otherwise a table of options described below
+        -- Abbreviation support
         abbreviations = {,
+            -- Set one of the following to true to enable abbreviations
+            compe = false, -- nvim-compe source
+            snippets = false, -- snippets.nvim source
+            enable = false, -- built-in expander
```
Julian, line 99: Can we call the option `builtin`, otherwise we've got `require'lean.abbreviations'.enable{enable = true}` as the lua API?

lua/lean/abbreviations.lua
```diff
@@ -33,18 +37,150 @@ local function snippets_nvim_enable(snippets, lean_abbreviations)
...
 
+local abbr_mark_ns = vim.api.nvim_create_namespace('leanAbbreviationMark')
+
+local function get_extmark_range(abbr_ns, id, buffer)
+  local row, col, details = unpack(
+    vim.api.nvim_buf_get_extmark_by_id(buffer or 0, abbr_ns, id, {details = true}))
+  return row, col, details and details.end_row, details and details.end_col
+end
+
+function M._clear_abbr_mark()
```
Julian, line 48: All these I think should be private (i.e. not exposed on the module) I think (though I'll be damned if I claimed I understood Lua scoping norms yet) -- but I think:

```suggestion
local function _clear_abbr_mark()
```

besides for the one you need to put on `_G`?

lua/lean/abbreviations.lua
```diff
@@ -33,18 +37,150 @@ local function snippets_nvim_enable(snippets, lean_abbreviations)
...
+
+  local tmp_extmark = vim.api.nvim_buf_set_extmark(0, abbr_mark_ns, row1, col1,
+    { end_line = row2, end_col = col2 })
+  local conv = function()
+    row1, col1, row2, col2 = get_extmark_range(abbr_mark_ns, tmp_extmark)
+    if not row1 or row1 ~= row2 then return end
+    vim.api.nvim_buf_del_extmark(0, abbr_mark_ns, tmp_extmark)
+    local text = vim.api.nvim_buf_get_lines(0, row1, row1+1, true)[1]:sub(col1 + 1, col2)
+    vim.api.nvim_buf_set_text(0, row1, col1, row2, col2, {M.convert_abbrev(text)})
+  end
```
Julian, line 148: We could add support for $CURSOR here right by setting the cursor to the right spot after checking if it's in the expanded value? Is that something you care about? Otherwise it's missing for compe too (#25) so we can always do it later.

scripts/minimal_init.lua
```diff
@@ -13,4 +13,8 @@ vim.api.nvim_exec([[
   runtime! plugin/plenary.vim
 ]], false)
 
-require('lean').setup{}
+require('lean').setup{
+  abbreviations = {
+    snippets = true,
```
Julian, line 18: It's a bit annoying that I still can't figure out a good way to avoid all the global state involved here, otherwise we'd be able to test the different implementations. But the neovim testing scene is still super immature, so oh well I suppose.

Julian (approved):
This is great, thanks again! Left a few minor comments, address whichever seem reasonable but otherwise lgtm!

lua/lean/abbreviations.lua
gebner: I've made all functions local that are not called from vimscript.  (For some reason I thought that if a function needs access to `M` it needs to be called `M.frob`.)

gebner:
Thanks for reviewing the PR!

I'll leave the `$CURSOR` abbreviations as todo.  They used to be implemented differently in vscode, and I honestly did not know about them.  Originally, they were called "hacky replacements" (in a comment in the code).  When you type `\{{` in vscode, it automatically adds closing braces and you get `\{{}}` with the cursor in the middle.  The idea is that this should be converted to `⦃⦄`.  The `$CURSOR` abbreviations are just a way to store this in the json file instead of a switch statement (they still get special treatment).  It's a bit less relevant now because `⦃⦄` is no longer a thing in Lean 4.

So I think implementing this in lean.nvim only makes sense if it works well with an auto-closing plugin (and I'm not even using one yet).  Otherwise you can just write `\c[...\c]` instead of `\{{}}...` (with about as many keystrokes). †

† I've just noticed that `\}}` doesn't trigger an abbreviation, so you can't use it for `⦄` with the built-in expander (write `\c]` as a workaround).  In my eyes that's a good thing though because `{\a}` immediately converts to `{α}` after typing the final `}`.

Julian:
Honestly I only use a $CURSOR-using thing for writing \<> instead of \<>, because the former involves dancing around the keyboard in uncomfortable ways. But I'm perfectly happy with it not being implemented yet. Thanks for the PR!

PR #40, opened 2021-06-07
Show term goals.

gebner:
Companion PR to leanprover/lean4#504.  There should be no merge conflicts with #35.

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -31,19 +31,28 @@ function M.update(infoview_bufnr)
     local params = vim.lsp.util.make_position_params()
     -- Shift forward by 1, since in vim it's easier to reach word
     -- boundaries in normal mode.
-    params.position.character = params.position.character + 1
-    return vim.lsp.buf_request(0, "$/lean/plainGoal", params, function(_, _, result)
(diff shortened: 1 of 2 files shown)


gebner:
Right, if you have `f a b` and put the cursor between `a` and `b`, then it shows the goal for `f a b`.  It only shows goals for terms that intersect the cursor position.  I'm not sure there's one single correct answer here.  Apparently you've expected the goal for `b`, but the goal for `a` would make just as much sense.

That's why I've added a range to the response, so that it's at least clear what subterm the server picked.  In another branch I have some code that highlights the range when you move your cursor over the "expected type" heading in the infoview: https://github.com/gebner/lean.nvim/tree/termgoal

> Without having any idea how much complexity it adds,

There is a helper function that allows you to get the smallest info node (i.e. smallest covering range satisfying some predicate).  I tried to keep the logic simple.  It was already awkward enough to exclude the `f` in `f a b` (so that you get the goal for `f a b` instead of `f`'s function type).

Julian:
Aha. That makes a lot of sense, and for some reason I had the opposite expectation in the test case that did `min a < min b` (that I'd see the whole term) so I can see things getting funky there. Thanks for explaining (and yeah lgtm again).

PR #42, opened 2021-06-09
Add :LeanPlainTermGoal in case someone wants inline-hover term goals.

Julian:
Demo:

[asciicast](https://asciinema.org/a/7gZzqLn1JfoykRy23EAL5iIDw)

### DIFF
--- a/lua/lean/lsp.lua
+++ b/lua/lean/lsp.lua
@@ -1,20 +1,27 @@
 local M = { handlers = {} }
 
+local function position_params()
+  -- Shift forward by 1, since in vim it's easier to reach word
+  -- boundaries in normal mode.
+  local params = vim.lsp.util.make_position_params()
+  params.position.character = params.position.character + 1
+  return params
+end
+
 function M.enable(opts)
   opts.commands = vim.tbl_extend("keep", opts.commands or {}, {
     LeanPlainGoal = {
-      function ()
-        local params = vim.lsp.util.make_position_params()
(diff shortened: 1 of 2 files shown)


lua/lean/lsp.lua
```diff
@@ -33,4 +40,18 @@ function M.handlers.plain_goal_handler (_, method, result, _, _, config)
...
+  config.focus_id = method
+  if not (result and result.goal) then
+    return
+  end
+  local markdown_lines = vim.lsp.util.convert_input_to_markdown_lines(result.goal)
+  markdown_lines = vim.lsp.util.trim_empty_lines(markdown_lines)
+  if vim.tbl_isempty(markdown_lines) then
+    return
+  end
+  return vim.lsp.util.open_floating_preview(markdown_lines, "leaninfo", config)
```
gebner, line 54: Why is the goal converted to markdown if the filetype is leaninfo?

gebner (commented):
This looks great!  There should probably be an entry in the README for the new command though.

lua/lean/lsp.lua
Julian: That's me bad-copypasting from the above. Will fix.

Julian:
The README needs some love (which Patrick pointed out as well) -- I had some local changes on a start there, but for now, just added two lines on the two current commands so they're at least mentioned (and fixed the unnecessary markdownizing).

PR #45, opened 2021-06-12
Make test timeouts customizable.

rish987:
I need this for my slow computer :*(


Julian:
Hah, nice, thanks!

Julian:
Probably these numbers are even too low for defaults, given that they short circuit, but I put them this low while I was debugging an annoying test failure.

So if we want to raise them we totally can (that was the one thing in the test changes you sent previously that I definitely did understand). But merged, thanks!

PR #48, opened 2021-06-12
Temporarily set LSP root directory search until neovim/nvim-lspconfig#958 is merged.

rish987:
It's been a few days, should I ping someone about merging neovim/nvim-lspconfig#958?


Julian:
We can yeah possibly try pinging in the Neovim gitter, mjlbach is usually responsive there.

PR #49, opened 2021-06-12
Internalize filetype detection.

rish987:
Also removes dependency on leanprover/lean.vim.

### DIFF
--- a/lua/lean/ft.lua
+++ b/lua/lean/ft.lua
@@ -1,28 +1,9 @@
 local M = {}
 
-local find_project_root = require('lspconfig.util').root_pattern('leanpkg.toml')
-
--- Ideally this obviously would use a TOML parser but yeah choosing to
--- do nasty things and not add the dependency for now.
-local _MARKER = '.*lean_version.*\".*:3.*'
-
-function M.init3()
-  pcall(vim.cmd, 'TSBufDisable highlight')  -- tree-sitter-lean is lean4-only
-  vim.b.lean3 = true
-end
-
-function M.is_lean3_project()
-  local project_root = find_project_root(vim.api.nvim_buf_get_name(0))
-  if not project_root then return false end
-  local result = vim.fn.readfile(project_root .. '/leanpkg.toml')
-  for _, line in ipairs(result) do
-    if line:match(_MARKER) then return true end
-  end
-  return false
-end
+local lean3 = require("lean.lean3")
 
 function M.detect()
-  if M.is_lean3_project() then M.init3() end
+  if lean3.is_lean3_project() then lean3.init() end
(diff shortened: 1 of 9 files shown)


rish987:
Filetype detection unit test implemented at: https://github.com/rish987/lean.nvim/tree/tests/internal_ftdetect.

BTW: I realize there are technically two parts to this PR, but it didn't feel right implementing filetype detection and continuing to recommend they install `lean.vim`, hence why these are bundled together.

lua/lean/ft.lua
```diff
@@ -0,0 +1,29 @@
+local M = {}
+
+local find_project_root = require('lspconfig.util').root_pattern('leanpkg.toml')
+
+-- Ideally this obviously would use a TOML parser but yeah choosing to
+-- do nasty things and not add the dependency for now.
+local _MARKER = '.*lean_version.*\".*:3.*'
+
+function M.init3()
```
Julian, line 9: I was trying to keep most of the Lean3 related things all together and in one file -- it's still far off, but that will make it easier to eventually hopefully just delete one file and only need one or two other lines of changes to get rid of it.

Can we put everything here except probably the `.detect()` function itself back in the lean3 file?

lua/lean/ft.lua
```diff
@@ -0,0 +1,29 @@
...
+  pcall(vim.cmd, 'TSBufDisable highlight')  -- tree-sitter-lean is lean4-only
+  vim.b.lean3 = true
```
Julian, line 11: I think we don't need this line anymore right?

Julian:
Nice, thanks again! Left one minor comment. Also look out for a collab invite, may as well have merge rights.

lua/lean/ft.lua
```diff
@@ -0,0 +1,10 @@
+local M = {}
+
+local lean3 = require("lean.lean3")
+
+function M.detect()
+  if lean3.is_lean3_project() then lean3.init() end
```
Julian, line 6: For me locally it seems like these lines need to be in the opposite order maybe? Probably because tree-sitter re-activates if the filetype is set (which maybe is a reason we should be using setfiletype instead of setting ft directly?)

But see this link: https://asciinema.org/a/3xYEDx2cUIv857hLXgRcZKIbv

The first time I open the file I get lean4 syntax highlighting (the white underlines are how my colorscheme indicates syntax errors), whereas afterwards if I reorder these two lines so that the .ft line is first, I get proper lean3 highlighting.

rish987: Yep that makes sense. Haven't used treesitter yet, so didn't notice this. Fixed.

Julian:
Cool, lgtm, merging, thanks again!

PR #50, opened 2021-06-13
Use 'minfixwidth to handle width setting, rather than manual futzing.

Julian:
`max_width` has been replaced with a set `width` setting.

Also bumping down the "recommended" size from 79 to 50, which is closer to what I've been using, but comments welcome if any of you two prefer the larger size.

For prior art, see how nvim-dap-ui does this:

https://github.com/rcarriga/nvim-dap-ui/blob/af1361c6ffac5285ca88e14d06dd5b949eabd682/lua/dapui/windows.lua#L55-L59

Besides removing code, this fixes the infoview getting unreadably small
in very small windows, and also not applying max_width correctly on
VimResized (though the latter would have been fixable), i.e. when the
whole vim gets resized.

(winfixheight presumably will be necessary/nice if/when we have horizontal infoviews.)

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -15,6 +15,7 @@ local _DEFAULT_WIN_OPTIONS = {
   number = false,
   relativenumber = false,
   spell = false,
+  winfixwidth = true,
   wrap = true,
 }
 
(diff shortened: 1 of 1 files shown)


rish987:
Awesome, this is much cleaner!

PR #52, opened 2021-06-13
Implement simultaneous lean3 and lean4 support with minimal lean4 syntax support.

rish987:
Pending neovim/nvim-lspconfig#958. (EDIT: merged)


Julian:
How about the Lean 4 syntax support, I don't see that here, no?

rish987:
Thought I'd leave that to the next PR. Though I can add it now if you'd prefer.

Julian:
Merging this though anyhow since at the minute, we're broken due to the filetype change in lspconfig, so better half-broken than broken. But yeah as-is I think this breaks syntax highlighting for Lean 4 without tree-sitter, so we should fix that ASAP.

rish987:
Yep on it.

PR #56, opened 2021-06-13
Add lean4 filetype detection tests.

rish987:
^^^


Julian:
More tests, hooray. Thanks.

PR #58, opened 2021-06-14
Extend unit tests to Lean 4 files where applicable.

rish987:
`clean_buffer()` now does an `it()` for `lean3` and `lean` buffers by default, and introduced `clean_buffer_ft()` to use for specific filetypes.


Julian:
Thanks! We'll need to [modify `.github/workflows/ci.yml` as well](https://github.com/Julian/lean.nvim/blob/main/.github/workflows/ci.yml#L13) to add a lean 4 version there and make this actually run on lean4 in CI, otherwise it's not actually running I think.

Also it makes me a bit uneasy to be honest to make `clean_buffer` do `it()` -- the latter is a "fundamental" part of the test framework, the former is a fixture-helper we wrote, so it seems a bit magic to make it do the latter to me. Someone reading the tests is likely to miss that that defines a test (and a bug may cause the test to not run at all, say).

Can we instead maybe just have something we pass into `it` that runs the 2 scenarios? E.g.:

```
it("does foo", lean4_and_3(clean_buffer("", [[
..))
```

?

Julian:
Ah wait.. what, that's already running properly in CI? How... /me thinks

rish987:
> Ah wait.. what, that's already running properly in CI? How... /me thinks

Good question - we're in the clear right now because we haven't tested LSP yet with standalone `lean` filetype buffers (using `clean_buffer()`), so the fact that `elan`'s default Lean version is 3 doesn't matter. And maybe we _never_ will need to test this? If we expect the Lean 4 use case to always be within a valid project tree, or expect `elan` to update to default to Lean 4 soon.

rish987:
I've updated `clean_buffer()` to not call `it()` anymore - however, this means that `lean3` and `lean` tests for a single `it()` are no longer separate test cases, so this could make the output harder to read (especially in failure). Will look into this in the future.

Julian:
> I've updated clean_buffer() to not call it() anymore - however, this means that lean3 and lean tests for a single it() are no longer separate test cases, so this could make the output harder to read (especially in failure). Will look into this in the future.

Ah.. yeah I see, that's annoying. I guess in other test frameworks now is when you use their parameterized test functionality but busted / plenary's busted may not have that (yet?)...

What you have now that explicitly lets you select a ft looks like an improvement though, so will merge, thanks! We can obviously improve further when we need to, but it's probably already nice that we can write an explicit lean4 test.

PR #61, opened 2021-06-15
Automatically generate vim documentation

Julian:
This PR is noisy because of a commit (870d2472e0c1d77c856407f4f218f6b0fc4c3773) that seems required to get proper tags generated for module names... which is unfortunate, but without it, the output has `M.load()` in it as the name of `abbreviations.load`, and I don't see how to get around that...

The real change is just 4114bdf951d494f7986b4807f11b81de65cb5b0e, which:

* Gives us CI that auto-regenerates the `doc/lean.txt` file
* Adds exactly 2 new docstrings just to try this out

If you want to see the output, run `:h lean` and you should now see some vim docs (or at least some empty space waiting for them...). Or obviously they're here: https://github.com/Julian/lean.nvim/blob/doc/doc/lean.txt which are generated from the Lua docstrings.

For info on how to write those magic Lua docstrings, see here: https://github.com/tjdevries/tree-sitter-lua/blob/6497ca7a3ff963e2fcb6bb3e39d7699280ae633e/HOWTO.md

Comments welcome of course. (And if this is annoying to review because of the commit noise, do let me know).

### DIFF
--- a/.github/workflows/docgen.yml
+++ b/.github/workflows/docgen.yml
@@ -1,0 +1,59 @@
+name: Generate documentation
+
+on:
+  push:
+    branches-ignore:
+      - master
+
+jobs:
+  build-sources:
+    name: Generate docs
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - run: date +%F > todays-date
+      - name: Restore cache for today's nightly.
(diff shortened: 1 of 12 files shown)


rish987 (approved):
These docs look great! I actually prefer using the full names over `M`.

Issue #60, opened 2021-06-15
Infoview doesn't render until second update


Julian:
When first opening a file, the infoview doesn't display the first state of the file (until a second update e.g. by moving the cursor).

Most likely the CursorHold triggers before the server is ready.

Demo below (for Lean 3):

[![asciicast](https://asciinema.org/a/tnRXxLvEaC7CO2LVV6scxb2IJ.svg)](https://asciinema.org/a/tnRXxLvEaC7CO2LVV6scxb2IJ)

rish987:
Should be resolved in Lean 4 by #85. Will investigate whether/how this is handled in `leanprover/vscode-lean` for Lean 3.

Julian:
This looks fixed for Lean 3 as well now:

https://user-images.githubusercontent.com/329822/136055079-0e0d9990-c2c6-4d83-bdcc-94a4f25e2079.mov

So thanks for somehow fixing it at some point :D

PR #62, opened 2021-06-15
Add more infoview tests.

rish987:
Tried to be exhaustive with this; let me know if I missed anything.

### DIFF
--- a/lua/tests/helpers.lua
+++ b/lua/tests/helpers.lua
@@ -5,8 +5,6 @@ local lean = require('lean')
 local api = vim.api
 local helpers = {_clean_buffer_counter = 1}
 
-local timeout = vim.env.LEAN_NVIM_TEST_TIMEOUT or 1000
-
 -- everything disabled by default to encourage unit testing
 local default_config = {
   treesitter = {
@@ -62,7 +60,7 @@ function helpers.clean_buffer(contents, callback)
 end
 
 function helpers.lsp_wait()
-  local succeeded, _ = vim.wait(timeout, vim.lsp.buf.server_ready)
+  local succeeded, _ = vim.wait(10000, vim.lsp.buf.server_ready)
   assert.message("LSP server was never ready.").True(succeeded)
 end
 
@@ -96,7 +94,7 @@ end
 
 --- Wait a few seconds for line diagnostics, erroring if none arrive.
 function helpers.wait_for_line_diagnostics()
-  local succeeded, _ = vim.wait(timeout * 2, function()
+  local succeeded, _ = vim.wait(5000, function()
     return not vim.tbl_isempty(vim.lsp.diagnostic.get_line_diagnostics())
   end)
   assert.message("Waited for line diagnostics but none came.").True(succeeded)
--- a/lua/tests/helpers.lua
+++ b/lua/tests/helpers.lua
@@ -61,22 +61,28 @@ function helpers.clean_buffer(contents, callback)
   end
 end
 
+function helpers.lsp_wait()
+  local succeeded, _ = vim.wait(timeout, vim.lsp.buf.server_ready)
+  assert.message("LSP server was never ready.").True(succeeded)
+end
+
 --- Create a clean Lean buffer of the given filetype with the given contents.
 --
 --  Waits for the LSP to be ready before proceeding with a given callback.
 --
 --  Yes c(lean) may be a double entendre, and no I don't feel bad.
-function helpers.clean_buffer_ft(ft, contents, callback)
+function helpers.clean_buffer_ft(ft, contents, callback, enter)
   return function()
     local bufnr = vim.api.nvim_create_buf(false, true)
     set_unique_name_so_we_always_have_a_separate_fake_file(bufnr)
+    -- apparently necessary to trigger BufWinEnter
+    if enter then vim.api.nvim_set_current_buf(bufnr) end
 
     api.nvim_buf_call(bufnr, function()
       require("lean.ft").set(ft)
       local this_lsp = ft == "lean" and lean.config.lsp or lean.config.lsp3
       if this_lsp.enable ~= false then
-        local succeeded, _ = vim.wait(timeout, vim.lsp.buf.server_ready)
-        assert.message("LSP server was never ready.").True(succeeded)
+        helpers.lsp_wait()
       end
 
       api.nvim_buf_set_lines(bufnr, 0, -1, false, vim.split(contents, '\n'))
--- a/lua/tests/helpers.lua
+++ b/lua/tests/helpers.lua
@@ -70,8 +70,8 @@ function helpers.clean_buffer(contents, callback)
   end
 end
 
-function helpers.lsp_wait()
-  local succeeded, _ = vim.wait(10000, vim.lsp.buf.server_ready)
+function helpers.wait_for_ready_lsp()
+  local succeeded, _ = vim.wait(20000, vim.lsp.buf.server_ready)
   assert.message("LSP server was never ready.").True(succeeded)
 end
 
@@ -91,7 +91,7 @@ function helpers.clean_buffer_ft(ft, contents, callback)
       require("lean.ft").set(ft)
       local this_lsp = ft == "lean" and lean.config.lsp or lean.config.lsp3
       if this_lsp.enable ~= false then
-        helpers.lsp_wait()
+        helpers.wait_for_ready_lsp()
       end
 
       api.nvim_buf_set_lines(bufnr, 0, -1, false, vim.split(contents, '\n'))
--- a/lua/tests/helpers.lua
+++ b/lua/tests/helpers.lua
@@ -31,6 +31,17 @@ function helpers.setup(config)
   require("lean").setup(vim.tbl_extend("keep", config, default_config))
 end
 
+local function has_all(_, arguments)
+  local text = arguments[1]
+  local expected = arguments[2]
+  for _, string in pairs(expected) do
+    assert.has_match(string, text, nil, true)
+  end
+  return true
+end
+
+assert:register("assertion", "has_all", has_all)
+
 --- Feed some keystrokes into the current buffer, replacing termcodes.
 function helpers.feed(text, feed_opts)
   feed_opts = feed_opts or 'n'
--- a/lua/tests/helpers.lua
+++ b/lua/tests/helpers.lua
@@ -80,12 +80,12 @@ end
 --  Waits for the LSP to be ready before proceeding with a given callback.
 --
 --  Yes c(lean) may be a double entendre, and no I don't feel bad.
-function helpers.clean_buffer_ft(ft, contents, callback, enter)
+function helpers.clean_buffer_ft(ft, contents, callback)
   return function()
     local bufnr = vim.api.nvim_create_buf(false, true)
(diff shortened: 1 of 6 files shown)


rish987:
(sorry for accidentally making this branch on upstream - was messing around with Github CLI)

lua/tests/infoview_lsp_spec.lua
```diff
@@ -0,0 +1,62 @@
...
+local helpers = require('tests.helpers')
+
+local function get_info_lines(infoview_info)
+  return table.concat(vim.api.nvim_buf_get_lines(infoview_info.bufnr, 0, -1, true), "\n")
+end
+
+local function infoview_lsp_test(pos, expected)
+    local infoview_info = infoview.open()
+    vim.api.nvim_win_set_cursor(0, pos)
+    infoview.update(infoview_info.bufnr)
```
rish987, line 11: Also tried replacing this update with `vim.wait(5000)`, but couldn't get `CursorHold` to trigger for some reason - though I do prefer it this way because it's more like a unit test. Probably best to test `CursorHold` separately (`TODO`).

Julian:
> (sorry for accidentally making this branch on upstream - was messing around with Github CLI)

Nah no big deal, either way's fine.

lua/tests/infoview_lsp_spec.lua
Julian: Yeah I couldn't get CursorHold to fire either when I tried this awhile ago... which is why I instead wrote an explicit test helper to crank (i.e. update) the infoview, and didn't depend on it happening on hold.

Julian:
Nice. Thanks!

Julian:
Nice. Thanks!

Something odd is happening in one of the tests I think though -- it happens both locally and I see in CI:

https://github.com/Julian/lean.nvim/runs/2834969873?check_suite_focus=true#step:10:102

See how it says there's an error? But it's clearly exiting with successful exit status... No idea what's up there. Any chance you see it?

rish987:
> See how it says there's an error? But it's clearly exiting with successful exit status... No idea what's up there. Any chance you see it?

Good catch - turns out we were't giving the Lean 4 server enough time to start up. I didn't notice this locally because I had manually set a larger timeout. Should be fixed now.

rish987:
The reason the error wasn't caught is because I put failing code inside of a `describe` block, rather than an `it` block. Fixed - though the real solution would be to use Busted's `setup` function. But it looks like `plenary.nvim` doesn't provide this yet.

lua/tests/helpers.lua
```diff
@@ -61,22 +59,28 @@ function helpers.clean_buffer(contents, callback)
   end
 end
 
+function helpers.lsp_wait()
```
Julian, line 62: Maybe `wait_for_ready_lsp` is a bit clearer on what we're waiting for?

lua/tests/helpers.lua
```diff
@@ -61,22 +59,28 @@ function helpers.clean_buffer(contents, callback)
...
 --  Waits for the LSP to be ready before proceeding with a given callback.
 --
 --  Yes c(lean) may be a double entendre, and no I don't feel bad.
-function helpers.clean_buffer_ft(ft, contents, callback)
+function helpers.clean_buffer_ft(ft, contents, callback, enter)
   return function()
     local bufnr = vim.api.nvim_create_buf(false, true)
     set_unique_name_so_we_always_have_a_separate_fake_file(bufnr)
+    -- apparently necessary to trigger BufWinEnter
+    if enter then vim.api.nvim_set_current_buf(bufnr) end
```
Julian, line 77: Are there tests that don't want to enter the buffer? Maybe we should always call this?

lua/tests/infoview_lsp_spec.lua
```diff
@@ -0,0 +1,62 @@
+local infoview = require('lean.infoview')
+local helpers = require('tests.helpers')
+
+local function get_info_lines(infoview_info)
+  return table.concat(vim.api.nvim_buf_get_lines(infoview_info.bufnr, 0, -1, true), "\n")
```
Julian, line 5: This seems useful enough to even be in the real code rather than just a test helper (as like `infoview.get_text` or something?)

lua/tests/infoview_lsp_spec.lua
```diff
@@ -0,0 +1,62 @@
...
+end
+
+local function infoview_lsp_test(pos, expected)
+    local infoview_info = infoview.open()
+    vim.api.nvim_win_set_cursor(0, pos)
+    infoview.update(infoview_info.bufnr)
+    local result, _ = vim.wait(5000, function()
+      for _, string in pairs(expected) do
+        local curr = get_info_lines(infoview_info)
+        if not curr:find(string) then return false end
```
Julian, line 15: It seems potentially a bit surprising that this does a partial match. Is it much more annoying to just return the contents after the wait, and let the caller use whichever assertion they want? And if luassert doesn't already have one (ugh) we add `assert.contains(foo, bar)`? Otherwise I guess fine as is, but yeah seems easy to forget this is not matching exactly.

lua/tests/infoview_lsp_spec.lua
```diff
@@ -0,0 +1,62 @@
...
+
+describe('infoview', function()
+  helpers.setup {
+    infoview = { enable = true },
+    lsp = { enable = true },
+    lsp3 = { enable = true },
+  }
+
+  it('lean 3', function()
+    vim.api.nvim_command("edit lua/tests/fixtures/example-lean3-project/test.lean")
```
Julian, line 31: Maybe we should be using [`before_each`](https://github.com/nvim-lua/plenary.nvim/blob/master/TESTS_README.md#a-simple-test)?

lua/tests/infoview_lsp_spec.lua
```diff
@@ -0,0 +1,62 @@
...
+    helpers.lsp_wait()
+    vim.wait(5000)
```
Julian, line 33: Seems like a long time to wait unconditionally here, what are we waiting for, the message to update?

Can we just do the assertion in the wait so it passes as soon as we find the match?

lua/tests/infoview_lsp_spec.lua
```diff
@@ -0,0 +1,62 @@
...
+
+    it('term state',
+    function(_)
+      infoview_lsp_test({3, 23}, {"⊢ ℕ"})
```
Julian, line 37: This test seems bare to me (i.e. the helper looks like it's doing too much, so it's not obvious what this test means). Probably related to the above on returning contents rather than doing the assertion inline. But thoughts welcome.

lua/tests/infoview_lsp_spec.lua
```diff
@@ -0,0 +1,62 @@
...
+    lsp = { enable = true },
+    lsp3 = { enable = true },
+  }
+
+  it('lean 3', function()
+    vim.api.nvim_command("edit lua/tests/fixtures/example-lean3-project/test.lean")
+    helpers.lsp_wait()
+    vim.wait(5000)
+
+    it('term state',
```
Julian, line 35:
```suggestion
    it('shows term state',
```

lua/tests/infoview_lsp_spec.lua
```diff
@@ -0,0 +1,62 @@
...
+    function(_)
+      infoview_lsp_test({3, 23}, {"⊢ ℕ"})
+    end)
+
+    it('tactic state',
```
Julian, line 40:
```suggestion
    it('shows tactic state',
```

Julian (commented):
Thanks again. Left some minor comments, but also it's tough to run this locally for me now, because I don't (and can't) use elan, so tests fail (I think because it can't simultaneously find lean 3 and lean 4, because both want to be called `lean`).

Maybe we can add an optional environment variable again that when set, explicitly tells the test suite where lean3 and lean4 are?

lua/tests/helpers.lua
```diff
@@ -31,6 +31,17 @@ function helpers.setup(config)
   require("lean").setup(vim.tbl_extend("keep", config, default_config))
 end
 
+local function has_all(_, arguments)
+  local text = arguments[1]
+  local expected = arguments[2]
+  for _, string in pairs(expected) do
+    assert.has_match(string, text, nil, true)
```
rish987, line 38: Did a nested assertion rather than returning `false` so that we could get `luassert`'s nicely formatted output on failure.

Julian:
OK I gave up and wrote myself a [poor man's elan](https://github.com/Julian/dotfiles/commit/2e86513de035bc1b2791dad27fcfa915f5b1dd19) so never mind that comment (though who knows, it may still be useful).

lua/tests/infoview_lsp_spec.lua
rish987: Not entirely sure what we're waiting for here, but if we request the `hover`/`plainGoal` data immediately without waiting it returns empty. I suspect we're waiting for the language server to finish a pass over the file, @gebner does that sound right?

You can try this out manually as well - when you first start things up, you'll notice that the infoview doesn't work for the first few seconds, regardless of if you're moving the cursor around (so it's not #60).

In Lean 4 there's the `processing...` diagnostic that is sent down the file to indicate progress - but rather than immediately implementing a check for that in these tests, I think we should wait until we've implemented that in the actual code (for a message in the infoview that indicates whether or not the server has processed up to the current line, similar to what is done in VSCode), which we can then use here.

Julian: So can we just condition on the infoview being nonempty here?

rish987: Think I fixed it as you've suggested - let me know what you think.

gebner: > Not entirely sure what we're waiting for here, but if we request the `hover`/`plainGoal` data immediately without waiting it returns empty. I suspect we're waiting for the language server to finish a pass over the file, @gebner does that sound right?

Yes, the requests always return the "current state".  If Lean hasn't processed that part of the file yet, you get an empty response.

You could try the (internal and completely standards-violating) `textDocument/waitForDiagnostics` request, which is supposed to block until Lean is finished.

Julian:
I combined the files into one which I think makes it a bit easier to find what's where, but other than that think lgtm, will merge as soon as CI finishes.

rish987:
Hmm are these still running in separate `nvim` instances? I put them in separate files I thought this was necessary for separate instances. Otherwise we'd be running `helpers.setup` more than once in a single instance, and these would be less like unit tests.

Julian:
AFAIK multiple files vs one files doesn't affect that really (there's some parallelization, but stuff can run in the same nvim).

So, it's always dirty (which of course is nonideal), and yes we're calling enable multiple times in the same nvim, but I think we could have been doing so previously as well.

rish987:
From nvim-lua/plenary.nvim:
> It will search for files with the pattern *_spec.lua and execute them in parallel in separate neovim instances.

Haven't looked into the code myself, but you're saying this line isn't entirely accurate?

Julian:
Never mind, it seems you're correct that plenary is mapping separate instances for each file: https://github.com/nvim-lua/plenary.nvim/blob/816d6c62bd020c0b29102be0890edb0c9e072373/lua/plenary/test_harness.lua#L70-L77

That's a bit unfortunate... but ok will revert.

Julian:
Merged, thanks again.

PR #64, opened 2021-06-19
Add basic LSP tests.

rish987:
Test file structure change seemed necessary to allow for sequential runs, as it looks like MacOS can't handle multiple simultaneous Lean 4 server instances.


lua/tests/fixtures/example-lean3-project/test.lean
```diff
@@ -1,5 +1,7 @@
 import test.test1
 
+set_option pp.notation false
```
Julian, line 3: Why do we need to disable this?

rish987, line 3: In some runs I noticed Lean 3 unexpectedly dropping notation, causing the test to fail, so I thought it was best to normalize this (haven't seen an issue with Lean 4 yet).

Julian (commented):
What kinds of bugs do you forsee these tests catching?

Don't they seem to just test LSP client <-> server interaction?

They're not even really testing code paths that are covered by the plugin, are they? They're just making LSP requests and asserting about how the server responds, so they seem more like they're testing the neovim LSP client and Lean LSP server.

rish987:
Yeah, all this is meant to test is that the LSP server is correctly enabled by the plugin. But on second thought the infoview tests already do that. But writing these was a good exercise for me 😆 

Closing...

Julian:
All good! Yeah I can imagine it wasn't the worst effort spent.

PR #65, opened 2021-06-19
Build Lean 3 fixture for tests.

rish987:
I figured since Lean 4 does this automatically it would be best to explicitly do this for Lean 3 as well, just to standardize things.


Julian:
Thanks!

PR #66, opened 2021-06-20
Separate Lean 3 and Lean 4 tests when using `clean_buffer`.

rish987:
Simplest way I could think to do this while maintaining the test structure of each `*_spec.lua` file.


Julian:
Merged, thanks again!

PR #67, opened 2021-06-22
Add infoview CursorHold(I) unit tests.

rish987:
Still haven't looked into how to trigger `CursorHold`/`CursorHoldI` from within our tests - however, I actually prefer doing it this way because we're only testing exactly what we need (the presence of the autocommands).


Julian (commented):
Thanks!

Issue #41, opened 2021-06-08
Backport the diagnostic and goal counting infoview sections to Lean3


Julian:
The [goal counting](https://github.com/Julian/lean.nvim/blob/a9ee9cde349eb2e150d7bacf1aca8abf6036b2cf/lua/lean/infoview.lua#L39-L48) and [diagnostic-showing](https://github.com/Julian/lean.nvim/blob/a9ee9cde349eb2e150d7bacf1aca8abf6036b2cf/lua/lean/infoview.lua#L58-L64) parts of the infoview work essentially unchanged on Lean3.

Just need a bit of refactoring to split apart the update function a bit better.

(The last item in #17 is obviously related, but this likely is a lot simpler to do for now).

Issue #17, opened 2021-05-23
More infoview tweaks


Julian:
These are a few other minor possible tweaks for the infoview. Some probably should be defaults, some configurable:

* [ ] automatically disable the infoview when it is `:q`'ed
* [ ] have one infoview per `:tab`
* [ ] only (or configurably) open the infoview only if there are no non-lean files open on the current tab
* [ ] split the infoview horizontally if there are already two vertical splits on the tab when it is opened
* [ ] define a (Lua) API for configuring how goals, diagnostics and headings are assembled into the buffer contents

PR #70, opened 2021-06-24
Backport diagnostic reporting to Lean 3.

Julian:
Does so by introducing infoview components (which in the future
possibly is a way to customize what's in the infoview display).

Closes: #41
Refs: #17

@gebner if you happen to have a spare moment I wouldn't mind a sanity check on whether [this crude splitting of Lean 3 goal state](https://github.com/Julian/lean.nvim/pull/70/files#diff-11f5cd6ffbb1e3a2d6e4a3948393a130292fd3d0f7e075249a20d6ebcafdb70cR19) is trivially incorrect in some way. It seems to work for the files / state I checked manually. If not no worries, can always revert later if it goes sideways.

### DIFF
--- a/doc/lean.txt
+++ b/doc/lean.txt
@@ -42,6 +42,11 @@ infoview.get_info_lines()                          *infoview.get_info_lines()*
 
 
 
+infoview.is_empty()                                      *infoview.is_empty()*
+    Is the infoview not showing anything?
+
+
+
 
 ================================================================================
                                                       *lean.infoview.components*
--- a/doc/lean.txt
+++ b/doc/lean.txt
@@ -44,6 +44,34 @@ infoview.get_info_lines()                          *infoview.get_info_lines()*
 
 
 ================================================================================
+                                                      *lean.infoview.components*
+
+Infoview components which can be assembled to show various information about
+the current Lean module or state.
+
+components.goal({goal})                                    *components.goal()*
+    The current (tactic) goal state.
+
+
+    Parameters: ~
+        {goal} (table)  a Lean4 `plainGoal` LSP response
+
+
+components.term_goal({term_goal})                     *components.term_goal()*
+    The current (term) goal state.
+
+
+    Parameters: ~
+        {term_goal} (table)  a Lean4 `plainTermGoal` LSP response
+
+
+components.diagnostics()                            *components.diagnostics()*
+    Diagnostic information for the current line from the Lean server.
+
+
+
+
+================================================================================
 sorry.fill()                                                    *sorry.fill()*
     Fill the current cursor position with `sorry`s to discharge all goals.
 
--- a/lua/lean/infoview/components.lua
+++ b/lua/lean/infoview/components.lua
@@ -1,0 +1,75 @@
+---@brief [[
+--- Infoview components which can be assembled to show various information
+--- about the current Lean module or state.
+---@brief ]]
+
+---@tag lean.infoview.components
+
+local DiagnosticSeverity = vim.lsp.protocol.DiagnosticSeverity
+local components = {}
+
+
+-- Format a heading.
+local function H(contents)
+  return string.format('▶ %s', contents)
+end
+
+--- The current (tactic) goal state.
+---@param goal table: a Lean4 `plainGoal` LSP response
+function components.goal(goal)
+  if type(goal) ~= "table" or not goal.goals then return {} end
+
+  local lines = {
+    #goal.goals == 0 and H('goals accomplished 🎉') or
+    #goal.goals == 1 and H('1 goal') or
+    H(string.format('%d goals', #goal.goals))
+  }
+
+  for _, each in pairs(goal.goals) do
+    vim.list_extend(lines, {''})
+    vim.list_extend(lines, vim.split(each, '\n', true))
+  end
+
+  return lines
+end
+
+--- The current (term) goal state.
+---@param term_goal table: a Lean4 `plainTermGoal` LSP response
+function components.term_goal(term_goal)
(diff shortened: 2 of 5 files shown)


rish987:
Thanks for cleaning up `infoview.lua`. This is interesting, I'm curious do you think it would be possible to do this by adding a new type of provider to [to `lean-language-server`](https://github.com/leanprover/lean-client-js/blob/520e4bb4b202c988ea496bb3994f9be164c7f853/lean-language-server/src/index.ts#L156) that does the parsing rather than doing that here?

lua/lean/lean3.lua
```diff
@@ -1,22 +1,52 @@
+local components = require('lean.infoview.components')
 local lean3 = {}
 
+-- Split a Lean 3 server response on goals.
+--
+-- Looks for ⊢, but ignores indented following lines for multi-line
```
gebner, line 6: That's the correct heuristic.  (⊢ at the beginning of a line until an empty line would work as well)

lua/tests/infoview/lsp_spec.lua
```diff
@@ -34,7 +34,7 @@ describe('infoview', function()
     it('shows term state',
     function(_)
       local text = infoview_lsp_update({3, 23})
-      assert.has_all(text, {"⊢ ℕ"})
+      assert.has_all(text, {"▶ goals accomplished 🎉"})
```
gebner, line 37: The result now means something completely different (previously 1 goal, now 0 goals).

syntax/leaninfo.vim
```diff
@@ -5,9 +5,9 @@ syn match leanInfoGoalVDash "^⊢"
 
 syn match leanInfoExpectedType "^▶ expected type.*"
 
-syn match leanInfoError "^▶.*: error:$"
-syn match leanInfoWarning "^▶.*: warning:$"
-syn match leanInfoInfo "^▶.*: information:$"
+syn match leanInfoError "^▶.*: Error:$"
```
gebner, line 8: I have to admit I personally prefer lowercase titles (because they don't stand out that much visually), but I'm willing to compromise on that.

However, the titles should stay consistent.  If "Error" is now capitalized, then "Expected type"  (and "Goals accomplished") should be capitalized as well.

lua/lean/infoview/components.lua
```diff
@@ -0,0 +1,84 @@
...
+--- Infoview components which can be assembled to show various information
+--- about the current Lean module or state.
+---@brief ]]
+
+---@tag lean.infoview.components
+
+local components = {}
+
+local _SEVERITY = vim.tbl_extend("error",
+  { [0] = "Other" },
```
gebner, line 11: If you want to reuse the built-in DiagnosticSeverity map, then you can drop the Other case.  I've only included it for completeness, but it's not used by Lean afaict.

lua/lean/infoview/components.lua
```diff
@@ -0,0 +1,84 @@
...
+
+  return lines
+end
+
+--- The combined tactic and/or term goal information from the Lean server.
+---@param goal table: a goal LSP response as returned by the Lean *4*
+---plainGoal request
+---@param term_goal table: a term goal LSP response as returned by the Lean *4*
+---plainTermGoal request
+function components.combined_goals(goal, term_goal)
```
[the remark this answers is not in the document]
Julian: Yep, reasonable. Done.

doc/lean.txt
```diff
@@ -43,6 +43,36 @@ infoview.get_info_lines()                          *infoview.get_info_lines()*
...
+
+Infoview components which can be assembled to show various information about
+the current Lean module or state.
+
+components.goal({goal})                                    *components.goal()*
+    The current (tactic) goal state.
+
+
+    Parameters: ~
+        {goal} (table)  a goal LSP response as returned by the Lean *4*
```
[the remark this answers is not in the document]
Julian: Yeah I was fighting with fixing that before my call... I can't figure it out at the minute, even though I see it's expected to work from tjdevries/tree-sitter-lua#14, so I pinged the maintainers.

For now I just shortened the docstring...

syntax/leaninfo.vim
Julian: Fair, agreed, reverted!

Julian:
> Thanks for cleaning up infoview.lua. This is interesting, I'm curious do you think it would be possible to do this by adding a new type of provider to lean-language-server that does the parsing rather than doing that here?
I think it'd be possible but it still wouldn't be 100% seamless, there's still some heuristic'ing because I think the upstream Lean 3 server message doesn't separate things out. And combined with knowing the lean-language-server repo isn't super active I thought it was easier to just shove here, though now Gabriel's merged the outstanding PR I had there.

lua/tests/infoview/lsp_spec.lua
Julian: Ech, yes, will have to check what's happening here, I changed it too mindlessly.

lua/lean/infoview/components.lua
Julian: Done.

lua/tests/infoview/lsp_spec.lua
Julian: Fixed, this was me not properly checking that the goal response had any goals in it...

Julian:
Dah, well done @rish987 -- I added `is_empty` precisely because I thought it'd fix it (and changed where you changed), but forgot the `table.concat`... *le facepalm*.

Appreciated!

rish987:
Fixed `infoview/lsp_spec.lua` LSP wait condition. LGTM at this point!

PR #71, opened 2021-06-25
Implement dynamically configurable infoview auto-open.

rish987:
The main refactor here is adding the ability to distinguish between "open", "closed", and "uninitialized". Should also set us up well for persistent infoview state features in the future.


README.rst
```diff
@@ -171,6 +171,8 @@ In e.g. your ``init.lua``:
         infoview = {
             -- Use a wider or narrower infoview.
             width = 50,
+            -- automatically open infoview on new tab?
```
Julian, line 174: This isn't on new tab right (I think you changed that since yesterday, that behavior was surprising to me, so I much prefer this way) -- it's now when entering a lean buffer, like it was before this PR?

If so is there really a reason someone would have `enable = truthy` and still `autoopen = false`? Can't those just be the same? Either the infoview is enabled (and then it autoopens) or not?

lua/tests/infoview/run_spec.lua
```diff
@@ -78,7 +89,7 @@ describe('infoview', function()
       assert.is_false(vim.api.nvim_win_is_valid(infoview_info.window))
       assert.is.equal(num_wins - 1, get_num_wins())
       assert.is.equal(win, vim.api.nvim_get_current_win())
-      assert.is_falsy(infoview.is_open())
+      assert.is_not.open_state()
```
Julian, line 92: This seems less readable than it was before to me.

README.rst
rish987: Yep, now `ensure_open()` only triggers on BufEnter. However, initialization only happens on new tab currently (until we get to per-window), which is where this option is considered, so the comment is accurate.

`enable = false` would disable infoview autocmds, though this does seem rather internal so we could make the behavior to always enable, and use the `enable` field to replace `autoopen`.

lua/tests/infoview/run_spec.lua
rish987: All this test did before was check that the internal state was updated (whether it was physically closed is done above). From that perspective, this would seem more readable, though let me know if you can think of a better name.

rish987: We could also do something like `assert.is_not.open_state(infoview)`.

README.rst
rish987: (see recent commit)

rish987: Oh I see what you're saying - yeah, we should say something along the lines of "automatically open infoview on entering lean buffer in new tab" instead

Julian: I guess I have to look at this PR a lot closer, I don't follow yet what it does really.

But my question was less about the name -- we have `infoview = false` already for "I don't want an infoview". What I was asking was what this option did beyond that

lua/tests/infoview/run_spec.lua
Julian: I don't follow why that's not `.is_open`?

`.is_open` checks some internal state of course, but it's (public) API -- an API that should mean "is the window open", so I'd expect it to encapsulate all the implementation details of what it means to be open or not. If the window was ever not open but that function returned true that'd be a bug right? So I'd expect whatever `open_state` checks to be the same thing `is_open` checks internally.

But I see this is related to closed and open not being opposites (which I think is quite confusing even though this is math-land where topologists roam). I guess I can't really understand the way it's done here yet until I follow what the PR is meant for as a whole.

README.rst
Julian: > Oh I see what you're saying - yeah, we should say something along the lines of "automatically open infoview on entering lean buffer in new tab" instead

Right now the comment makes sense to me at least, but I'm skeptical someone will ever want `infoview = not false` and `autoopen = false`, or do you see a scenario for that?

rish987:
Sorry I wasn't clear about this, will be better about creating/linking issues in the future. The purpose of this PR is to pass [this test case](https://github.com/Julian/lean.nvim/blob/56ec5a746e61a89950f33e5843532eac39f1c5ef/lua/tests/infoview/run_spec.lua#L183).

README.rst
rish987: This is for someone who's using A LOT of tabs/windows, so they may want to periodically control whether their new tabs/windows automatically open infoviews. Perhaps a bit rare, though I figured since the refactor (intended for #74) enables this anyways we might as well expose it.

lua/tests/infoview/run_spec.lua
rish987: There's a new state for the infoview - uninitialized. This check is the equivalent of:
```
assert.is_truthy(infoview.is_closed())
assert.is_falsy(infoview.is_open())
```

Let me know if you prefer it this way for readability.

Julian: I prefer we make that not possible!

But I'm going to review this in a few hours.

But I would much prefer closed to be a superset of uninitialized, `is_closed` to just be `not is_opened`, and end-users never see uninitialized.

Julian:
Aha. Thanks, the clarification helps definitely on what this is trying to do!

Will spend a bit of time with it.

lua/tests/infoview/run_spec.lua
rish987: Yeah it's actually not possible from a user's perspective, in that they'll never see an uninitialized infoview. So it's more of a sanity check for us.

rish987: Also I don't think we can erase the distinction between closed and uninitialized - closed means explicitly closed by the user, so don't open. Uninitialized means open only if they've enabled auto-open.

Julian:
Quick update here -- I [have a branch with a medium-sized refactor](https://github.com/Julian/lean.nvim/tree/oop) that personally I'm a lot more comfortable following, and which addresses some of what I was trying to get across (it essentially removes the uninitialized state, makes it impossible for a buffer to be both closed and open, removes lots of intertwined conditionals, handles set_autoopen by just setting or unsetting an autocmd, few other things).

Have a look and let me know what you think. I *believe* if I follow what you were saying previously about persistent stuff that that should be equally possible even after that branch.

There are still a few more things I want to address, but I think the only major one is a [test fails in CI](https://github.com/Julian/lean.nvim/runs/2960103902?check_suite_focus=true#step:11:179) that never fails locally, and I don't see why, so I'll have to diagnose what's different between my machine and the CI setup again...

Other than that one test, all the others pass essentially without modification from what we had previously (and in this PR).

rish987:
Awesome!!! Will check it out today. In general OOP sounds like a very good idea for infoviews.

lua/lean/infoview.lua
```diff
@@ -196,28 +192,29 @@ function infoview.enable(opts)
...
     autocmd FileType lean lua require'lean.infoview'.make_buffer_focusable()
   ]])
-  infoview.set_autoopen(options.autoopen)
 end
 
 --- Configure the infoview to update when this buffer is active.
 function infoview.make_buffer_focusable()
+  -- WinEnter is necessary for the edge case where you have
+  -- a file open in a tab with an infoview and move to a
+  -- new window in a new tab with that same file but no infoview
```
rish987, line 201: > I'm still not 100% sure why we focus windows on both WinEnter and BufEnter events -- are both needed? The tests seem to think so, but are we calling focus twice in some cases?

See here.

lua/lean/infoview.lua
```diff
@@ -61,42 +69,51 @@ function Infoview:new(id, width)
...
+  self.window = nil
+  self.is_open = false
+
+  self:focus_on_current_buffer()
 end
 
 --- Toggle this infoview being open.
 function Infoview:toggle()
-  self:close()
+  if self.is_open then self:close() else self:open() end
```
rish987, line 103: > Consider removing Infoview/ClosedInfoview.is_open or replacing with a call to vim.fn.win_findbuf or filtering on window names? Right now only tests access this attribute (essentially to differentiate between the two), but it'd seem better if those tests instead actually look to see whether there's an infoview window visible.

We're now using this field here, as well as a couple of other places. `vim.fn.win_findbuf` is a good idea as an additional testing criterion, but I think it wouldn't hurt to leave in the `is_open` check just as a quick internal sanity check for us.

rish987:
> Split the commit up

Good call, though I'm not so sure about the extent to which a refactor like this can be made to look incremental at any stage - maybe it's just a matter of reorganizing?

rish987:
> Issues which seem present on current `main` as well

Both of these I think deserve their own issues,

rish987:
@Julian Thank you very much for the refactor idea! I've fixed and simplified your branch and I'm pretty certain now that it matches your intentions, but let me know if I missed anything. Also tried to address your checklist (see above comments).

lua/lean/infoview.lua
Julian: Sounds good to me!

lua/lean/infoview.lua
Julian: Aha, thanks!

Julian (commented):
Awesome, thanks well done again...

Think we merge at this point. I'll file some issues for the other stuff but yeah glad we managed to get somewhere here!

Julian:
> > Split the commit up
> 
> Good call, though I'm not so sure about the extent to which a refactor like this can be made to look incremental at any stage - maybe it's just a matter of reorganizing?

This was a scribble from an earlier point -- yeah I could reorganize the commit into a few separate ones but at this point given you tidied the code up it's just gonna stay as is which is fine.

PR #72, opened 2021-06-25
Remove unnecessary line separator when no tactic state.

rish987:
Made this a PR just to make sure there wasn't a specific reason you did this. Also enables the `No info found.` message in Lean 4 and prevents a (very rare) `infoview/lsp_spec.lua` test failure.


Julian:
Thanks! No there definitely was not a reason for it, so thanks for fixing, though I did notice newlines were in need of some additional work (which I was planning on later) -- e.g. there's still an extra line at the top of the infoview when there's no goal or tactic state but there are diagnostics.

At some point I think we want the lua equivalent of `"\n".join(all the nonempty components that are enabled)` essentially.

But yeah thanks for fixing!

PR #73, opened 2021-06-26
Increase the default memory limit in Lean3.

Julian:
Matches VSCode's default found here:

https://github.com/leanprover/vscode-lean/blob/ca764c60acbfaeaca55a681bc0be65f84b0cae44/package.json#L32-L35

Possible further improvements here (which I think are slightly lower priority, but comments welcome) are to replicate something more like the VSCode settings where we have `opts.memory_limit` (and `opts.timeout`, etc.) and use those to assemble the CLI.

Or similarly to extend this to Lean4.

But really the main current issue is that the defaults aren't very usable, so I wanted to fix that ASAP.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -187,7 +187,12 @@ In e.g. your ``init.lua``:
 
         -- Lean 3
         lsp3 = {
-            on_attach = on_attach
+            on_attach = on_attach,
+
+            -- Additional options not passed directly to lspconfig
+            --
+            -- Extra CLI arguments to pass to `lean-language-server`
+            extra_argv = { "-M", "4096" }
         },
     }
 
--- a/lua/lean/lean3.lua
+++ b/lua/lean/lean3.lua
@@ -1,5 +1,5 @@
 local components = require('lean.infoview.components')
-local lean3 = {}
+local lean3 = { lsp = {} }
 
 -- Split a Lean 3 server response on goals.
 --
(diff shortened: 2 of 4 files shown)


lua/lean/lean3.lua
```diff
@@ -18,11 +18,30 @@ local lean3 = {}
...
+---@field extra_argv table: additional command line options to pass to `lean-language-server`
+function lean3.lsp.enable(opts)
+  require('lspconfig.lean3ls')
+  opts.cmd = opts.cmd or vim.deepcopy(
+    require('lspconfig/configs').lean3ls.document_config.default_config.cmd
+  )
+  local extra_argv = opts.extra_argv or { "-M", "4096" }
+  table.insert(opts.cmd, '--')
+  vim.list_extend(opts.cmd, extra_argv)
+  opts.extra_argv = nil
```
rish987, line 40:
```suggestion
  opts.cmd = opts.cmd or { "lean-language-server", "--stdio", "-M", "4096" }
```
Feel free to make this the default [on upstream](https://github.com/neovim/nvim-lspconfig/blob/48e47b4d26e45977d27c06dc1bc116416ca9276d/lua/lspconfig/lean3ls.lua#L6) as well (at which point this line can be removed).

README.rst
```diff
@@ -187,7 +187,12 @@ In e.g. your ``init.lua``:
 
         -- Lean 3
         lsp3 = {
-            on_attach = on_attach
+            on_attach = on_attach,
+
+            -- Additional options not passed directly to lspconfig
+            --
+            -- Extra CLI arguments to pass to `lean-language-server`
+            extra_argv = { "-M", "4096" }
```
rish987, line 195:
```suggestion
            -- This is the default if unspecified.
            cmd = { "lean-language-server", "--stdio", "-M", "4096" }
```
IMO probably best to avoid introducing new fields to the config passed directly to `nvim-lspconfig` if possible (see below).

rish987:
Just a couple minor comments, but otherwise LGTM. My only concern with the way things are currently is if someone actually used `cmd` as `nvim-lspconfig` intends for them to, with `cmd = { "lean-language-server", "--stdio", "--", "-M", "4096" }`, without using `extra_args`, their command would get an extra `"--", "-M", "4096"` appended.

Julian:
Yesssss thanks I was lazy yesterday :)

Moved upstream instead, neovim/nvim-lspconfig#1000.

rish987:
OMG PR number 1000 !?!? 👀 👀 👀

Issue #78, opened 2021-06-30
Outdated infoview updates can override the most recent one.


rish987:
See here: [![asciicast](https://asciinema.org/a/0obagpgwvSApsVaNV6jZ3bGP0.svg)](https://asciinema.org/a/0obagpgwvSApsVaNV6jZ3bGP0)

The solution should be similar to that of #77, also cancelling all requests at the beginning of every `update()` call so only the most recent one survives.

rish987:
Attempted fix at rish987/lean.nvim@f462dda. However, it doesn't work because the returned cancellation function from `vim.lsp.buf_request()` seemingly does nothing. You can even try this in `lsp.lua` (rish987/lean.nvim@3e2e21d):,
```lua
  -- Fetch goal state information from the server.                                                                                                                                                                                             
  function lsp.plain_goal(bufnr, handler)                                                                                                                                                                                                      
    -- Shift forward by 1, since in vim it's easier to reach word                                                                                                                                                                              
    -- boundaries in normal mode.                                                                                                                                                                                                              
    local params = vim.lsp.util.make_position_params()                                                                                                                                                                                         
    params.position.character = params.position.character + 1                                                                                                                                                                                  
   local id_map, cancel = vim.lsp.buf_request(bufnr, "$/lean/plainGoal", params, handler)                                                                                                         
    cancel()                                                                                                                                                                                                                                   
    -- This doesn't work either:                                                                                                                                                                                                               
    --                                                                                                                                                                                                                                         
    --for client_id, req_id in pairs(id_map) do                                                                                                                                                                                                
    --  vim.lsp.get_client_by_id(client_id).cancel_request(req_id)                                                                                                                                                                             
    --end                                                                                                                                                                                                                                      
  end 
```
and everything still passes.

Any ideas? Could this be a neovim bug? Will try to put together an MWE tomorrow.

rish987:
(^ This seems to have something to do with the cancel function sending a `cancelRequest` to the server, rather than just cancelling the callback, which is what I assumed it does. And it looks like the Lean servers don't do anything on `cancelRequest`? Investigating...)

PR #79, opened 2021-07-01
Cancel infoview `update()` requests where necessary.

### LINKED ISSUE #77: Closing infoview after triggering `update()` but before callback results in `Invalid buffer id:...` error.
Because when an infoview is closed, its buffer is deleted. The solution seems to be to cancel all pending requests on infoview close.

rish987:
This is a bit of a workaround instead of using the cancellation function returned by `vim.lsp.buf_request()` since it looks like the Lean servers weren't responding [to `cancelRequest`](https://github.com/neovim/neovim/blob/f35a5f2efc32b508ff5fc7d55b31a5bb7facfa17/runtime/lua/vim/lsp.lua#L958) with a [cancellation acknowledgement](https://github.com/neovim/neovim/blob/f35a5f2efc32b508ff5fc7d55b31a5bb7facfa17/runtime/lua/vim/lsp/rpc.lua#L538), which would normally [clear the callback](https://github.com/neovim/neovim/blob/f35a5f2efc32b508ff5fc7d55b31a5bb7facfa17/runtime/lua/vim/lsp/rpc.lua#L552). @gebner any plans to support this in the future?

Also #78 proved not so simple to deterministically reproduce, so leaving out a unit test for this for now.

Closes #77 and #78.

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -26,9 +26,38 @@ local function get_idx()
   return vim.api.nvim_win_get_tabpage(0)
 end
 
+function infoview.update_request(idx)
+  return function(bufnr, method, params, handler)
+    local metadata = {enable = true}
+    local map, cancel_func = vim.lsp.buf_request(bufnr, method, params, function(...)
+      if not metadata.enable then return false end
+      handler(...)
+      end)
+    vim.list_extend(infoview._infoviews[idx].req_metas, {metadata})
+    return map, cancel_func
+  end
+end
+
+function infoview.cancel_all(idx)
+  idx = idx or get_idx()
+
+  -- TODO: this shouldn't be necessary; the reason it currently is is that we
+  -- don't remove the LeanInfoviewClose augroup on _teardown(), so on close()
+  -- _teardown() is called twice
+  if infoview._infoviews[idx] == nil then return end
+
+  for _, req_meta in pairs(infoview._infoviews[idx].req_metas) do
+    req_meta.enable = false
+  end
+  infoview._infoviews[idx].req_metas = {}
+end
+
 function infoview.update()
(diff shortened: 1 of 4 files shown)


gebner:
> any plans to support this in the future?

Oh, this is indeed a bug.  Cancellation requests require a response per https://microsoft.github.io/language-server-protocol/specification.html#cancelRequest

Note that you cannot *rely* on requests being cancelled.  The server is allowed to ignore cancellation requests.  It is also allowed to return incorrect (i.e., partial) results.  (In some cases ignoring cancellation requests is even unavoidable, for example if the original request has already been finished when the cancellation request was received.)

rish987:
I see! So in that case it's probably best to leave things as they are here (or similar), even when/if cancellation requests are implemented. Don't know much about the Lean internals yet, but perhaps this is worth implementing if it can help reduce the load on the server? Thanks for the info.

rish987:
(putting on hold for a bit anticipating OOP refactor from #71)

gebner:
> Don't know much about the Lean internals yet, but perhaps this is worth implementing if it can help reduce the load on the server?

Yes, I'll report a bug (or maybe make a PR right away) when I have time.

rish987:
@Julian Let me know what you think of the (smaller-scale) refactor to make working with infoviews a bit more like using React. This should put us at the cusp of implementing pinnable/pausable messages.

rish987:
Okay on second thought, after brainstorming how to do #84 (probably using OOP for infoview messages), `set_state` probably won't be so useful. Seems best to keep it simple and just call `render()` explicitly when needed. Will update soon

rish987:
I've gotten pretty comfortable with the tick logic at this point, so I think I'll finally close this...

PR #80, opened 2021-07-01
Automatically upload coverage reports on CI builds

Julian:
First build here: https://codecov.io/gh/Julian/lean.nvim/tree/6b1fcec697c17db09dd7fbbc887b2b6673a4d1e6/lua

Obviously this is nice to have, but it's particularly nice I think because plenary's test runner is fragile -- while I was working on the infoview refactor I straight up broke `set_autoopen`, but rather than failing, tests were silently skipped by plenary because they failed in the `describe` block.

With coverage, we can at least find unexecuted test lines.

(Probably the codecov settings will need some tweaking, I never remember quite what the settings mean)

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,67 @@
+name: Coverage
+
+on: [push, pull_request]
+
+jobs:
+  coverage:
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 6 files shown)


rish987:
This looks really cool, thanks! Though it is all new to me, will look in greater detail soon.

Issue #84, opened 2021-07-05
Support pinnable infoview messages


rish987:
Similar to VSCode, it should be possible to pin/unpin infoview messages.

VSCode seems to support "live" messages that are [automatically updated](https://github.com/leanprover/vscode-lean/blob/0d4e372533990ef1a73e5674f5a095af4502cb80/infoview/info.tsx#L128) as the surrounding code changes. To me this seems a bit unexpected, and I would prefer to just have pinned messages be permanently "paused" and only update when specifically requested. Not to mention this would make the code simpler. Obviously if someone wants to simulate live updates this could be done with autocmds. Though keep in mind I don't have as much experience writing Lean code myself as reading it, so do let me know if there are other opinions.

rish987:
Update: see neovim/neovim#15276 -- basically I think we should closely follow the way things are done at VSCode, which is to [update pin positions on `textDocument/didChange`](https://github.com/leanprover-community/vscode-lean4/blob/d5f1926be34e73af23d551be779a5b98e721e0ba/lean4-infoview/src/infoview/infos.tsx#L14), rather than doing things ad-hoc with `nvim_buf_attach()` (the analogue of which was the case in VSCode before leanprover-community/vscode-lean4#30). This will guarantee that the call to `infoview.__update()` following updating the pins will come after actually having notified the server of the buffer changes.

PR #87, opened 2021-07-06
Add lean.current_search_paths for finding the current library paths

Julian:
Demo (of using it with telescope):

https://github.com/Julian/lean.nvim/wiki/Configuring-&-Extending#live-grep


gebner:
> The implementation here is really trivial for someone to do themselves, it's one line with telescope.

In that case, I think it would be more helpful to advertise an example configuration for telescope in the readme.  Live grep is also super useful outside of lean projects.

The lean.finder module doesn't seem to provide much convenience either.  The configuration you have in your init.vim seems to be longer than the equivalent telescope configuration.

> We may extend functionality here at some point.

What might make this functionality more interesting is if it picked up the directories automatically.  For lean 3, we can parse the output of `lean -p`.  In lean 4, this is unfortunately a bit more complicated.

Julian:
That's fair feedback, and likely a good idea. Will have a look at what's involved for the Lean 4 case.

(Though I apparently pushed this to main inadvertently, so will have to back it out).

gebner:
> Will have a look at what's involved for the Lean 4 case.

IIRC there's no nice way to get the paths.  You can get the paths to the source directories for the current package with `leanpkg print-paths` (the second lines is a :-separated list of directories).  This does not include the source directory for the core library.  The server does something like the following to figure it out:
```bash
echo 'def main : IO UInt32 := do IO.println <| <- IO.FS.realPath <| (<- IO.appDir) / ".." / "lib" / "lean" / "src"; pure 0' | lean --stdin --run
```

Julian:
>  The server does something like the following to figure it out

`lean --print-libdir` looks like it's [roughly that now](https://github.com/leanprover/lean4/commit/4ed66cae3e47dd186a47c83e237ace6bf559e2a6). It doesn't restrict to just the src directory but including the rest of the implementation seems possibly nice anyhow to me.

Will rewire to use that and see what you think.

lua/tests/current_search_paths_spec.lua
```diff
@@ -0,0 +1,24 @@
...
+require('tests.helpers').setup {
+  lsp = { enable = true },
+  lsp3 = { enable = true },
+}
+describe('lean.current_search_paths', function()
+  describe('lean3', function()
+    it('returns the current search paths', function()
+      vim.api.nvim_command("edit lua/tests/fixtures/example-lean3-project/test.lean")
+      helpers.wait_for_ready_lsp()
+      assert.is.truthy(lean.current_search_paths())
```
rish987, line 13: Could we have this check that the particular path returned is correct? And can we also have a test for nested files?

Julian: Yes! Good points, this was waiting for neovim/neovim#15058 to be fixed but will likely get back to it today or tomorrow.

Julian:
OK, this is ready for rereview!

lua/lean/init.lua
```diff
@@ -73,4 +70,36 @@ function lean.use_suggested_mappings(buffer_local)
...
+function lean.current_search_paths()
+  local paths
+
+  if vim.bo.ft == "lean3" then
+    paths = require'lean.lean3'.__current_search_paths()
+  else
+    local root = vim.lsp.buf.list_workspace_folders()[1]
+    paths = vim.tbl_map(
+      function(path) return root .. '/' .. path end,
+      subprocess_check_output({"leanpkg", "print-paths"}, { cwd = root })
```
gebner, line 91: This should only take the second line, and split it by `:`.  This is an example output:
```
$ leanpkg print-paths
././build:build/deps/Qq/./build
./.:build/deps/Qq/.
```

The directories with source files are `./.` and `build/deps/Qq/.`

Julian: Aha, sorry, missed that comment. The first line is some unrelated information?

lua/lean/_util.lua
```diff
@@ -23,4 +24,46 @@ function M.set_augroup(name, autocmds, buffer)
...
+  local lines, stderr_lines
+
+  opts = vim.tbl_extend(
+    "keep", opts or {}, {
+      stdout_buffered = true,
+      stderr_buffered = true,
+      on_stdout = function(_, data) lines = data end,
+      on_stderr = function(_, data) stderr_lines = data end
+    }
+  )
```
rish987, line 43: [Plenary's `Job` module](https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/job.lua) abstracts away a lot of these lower level details, perhaps there's an argument for using that instead of "manually" spawning a process with handlers?

Esp. relevant here are [`Job:result()`](https://github.com/nvim-lua/plenary.nvim/blob/8bae2c1fadc9ed5bfcfb5ecbd0c0c4d7d40cb974/lua/plenary/job.lua#L453), [`Job:stderr_result()`](https://github.com/nvim-lua/plenary.nvim/blob/8bae2c1fadc9ed5bfcfb5ecbd0c0c4d7d40cb974/lua/plenary/job.lua#L458), and [`Job:wait()`](https://github.com/nvim-lua/plenary.nvim/blob/8bae2c1fadc9ed5bfcfb5ecbd0c0c4d7d40cb974/lua/plenary/job.lua#L468).

Julian: It'd add a runtime dep for users (instead of just a testing one).

It's definitely annoying to need to NIH this though, I was complaining about it a bit.

But happy to go either way.

gebner: Plenary also has a nice async library with wrappers for LSP that would be nice to use instead of the callback hell.

lua/lean/init.lua
gebner: I think the first line is the directories where the *.olean files are stored.

lua/lean/_util.lua
rish987: Good point, IMO since our user-base seems like it still is pretty much still just us (and will be until we implement competitive features like #84) I'd say it's best to add the dependency now - I can work later today on refactoring this to use `Job`.

Julian: OK, sounds good to me! And thanks that'd be great, gonna probably hit a beer for the evening so feel free to play around, otherwise I'll address in the morning.

lua/lean/init.lua
```diff
@@ -73,4 +70,36 @@ function lean.use_suggested_mappings(buffer_local)
...
+    )
+    vim.list_extend(
+      paths,
+      subprocess_check_output({"lean", "--print-libdir"}, { cwd = root })
```
rish987, line 95: This I think could add a lot of noise to the search - perhaps we should make it configurable whether to include the standard library?

rish987: We can also more generally start thinking about how to allow someone to choose what dependencies are included in the search (after splitting as suggested by Gabriel above)

lua/tests/current_search_paths_spec.lua
```diff
@@ -0,0 +1,49 @@
...
+  lsp = { enable = true },
+  lsp3 = { enable = true },
+}
+describe('lean.current_search_paths', function()
+  for kind, path in pairs{
+    ["existing"] = fixtures.lean_project.some_existing_file,
+    ["nested existing"] = fixtures.lean_project.some_nested_existing_file,
+    ["nonexisting"] = fixtures.lean_project.some_nonexisting_file,
+    ["nested nonexisting"] = fixtures.lean_project.some_nested_nonexisting_file
+  } do
```
rish987, line 15: Good idea putting these paths strings in a helper rather than duplicating. I think we can define a custom iterator for this in the helper as well (which we can also use in `ft_spec.lua`), will do.

lua/lean/init.lua
rish987: Actually, I have an idea for some upstream changes to `telescope.nvim`, which should allow for this and also make it so that we can configure integration here (rather then asking users to do this in their dotfiles). Will make an issue with more details tomorrow.

lua/tests/current_search_paths_spec.lua
```diff
@@ -0,0 +1,39 @@
...
+}
+describe('lean.current_search_paths', function()
+  for kind, path in unpack(fixtures.lean_project.files_it) do
+    it(string.format('returns the paths for %s files', kind), function()
+      vim.api.nvim_command('edit ' .. path)
+      helpers.wait_for_ready_lsp()
+
+      local paths = lean.current_search_paths()
+      -- via its leanpkg.path:
+      assert.has_all(
```
rish987, line 17: We should probably make these tests more specific (i.e. using [exact match with `assert.has_match()`](https://github.com/Julian/lean.nvim/blob/d534ad4bf3e545d5be6e16b7746eff6304421119/lua/tests/helpers.lua#L113)) to catch the bug Gabriel noticed above.

Julian: It's hard to do that I think without doing something more complex than the actual implementation, given the paths are absolute (intentionally) and the Lean stdlib location will even vary by machine -- do you see an obvious way to do it?

To fix the bug above I need to basically add a second directory (because right now no colons are in the output for the test case) but yeah let me know what you think.

lua/lean/init.lua
Julian: Fixed I believe.

lua/lean/init.lua
Julian: I guess it depends how much overlap what you type has with stuff in the stdlib -- I could certainly believe it'd be useful to not show it though. But yeah maybe in a followup? There's other stuff I want to hit.

gebner: Searching the stdlib is the only reason why I think this feature should be in lean.nvim at all.  I already have `<leader>fg` set up to do live grep in any project, Lean or not (this is a one-liner directly copied from the telescope.nvim readme).

In mathlib, the boundary between what is in the core library and what is in mathlib is not very clearly defined.  Some basic definitions are in mathlib, others in core.  So if you're already searching through all of mathlib, you should probably search through the core library as well.

lua/tests/current_search_paths_spec.lua
rish987: No problem, I think what you've just done (checking the suffix using `\n`) is already good enough.

rish987:
@gebner why is it that for Lean 3, `lean --path` also returns a path to a nonexistent `../library` directory (see recent failure)?

```
$ lean --path
{
  "is_user_leanpkg_path": false,
  "leanpkg_path_file": "/home/rishhvaishnav/.vim/pack/bundle/start/lean.nvim/lua/tests/fixtures/example-lean3-project/l
eanpkg.path",
  "path": [
    "/home/rishhvaishnav/.elan/toolchains/leanprover-community--lean---3.31.0/bin/../library",
    "/home/rishhvaishnav/.elan/toolchains/leanprover-community--lean---3.31.0/bin/../lib/lean/library",
    "/home/rishhvaishnav/.vim/pack/bundle/start/lean.nvim/lua/tests/fixtures/example-lean3-project/./."
  ]
}
```

gebner:
> why is it that for Lean 3, lean --path also returns a path to a nonexistent ../library directory (see recent failure)?

Because if you use lean directly from the build directory, you only have `~/lean/bin/../library` (and not `~/lean/bin/../lib/lean/library`).

rish987:
@Julian see recent few changes and let me know if you have any comments, otherwise LGTM at this point.

Julian:
Nice, thank you!

PR #90, opened 2021-07-08
Render infoview contents on update iff `is_open`.

### LINKED ISSUE #77: Closing infoview after triggering `update()` but before callback results in `Invalid buffer id:...` error.

rish987:
Separated from #79. Closes #77.

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -118,8 +118,30 @@ function Infoview:focus_on_current_buffer()
   end
 end
 
---- Update this infoview's contents given the new set of lines.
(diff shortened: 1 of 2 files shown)


Julian (approved):
lgtm, though at some point I'd like to bring back the interface-driven structure I had previously (with `ClosedInfoview`), which removes all these if statements littered in each method (e.g. here, a closed infoview's update method should obviously do nothing).

But for now this looks good, thanks again!

PR #99, opened 2021-07-15
Allow for specifying a single test file to run.

rish987:
Roughly follows the way this is done [in neovim's `busted` tests](https://github.com/neovim/neovim/blob/682247b52e76180e4c49b794c6cc2548bfd32d4c/cmake/RunTests.cmake#L22), that is to run a test for only a specific file you can do e.g.:

```
$ TEST_FILE=lua/tests/sorry_spec.lua make test
```


Julian:
Great!

PR #102, opened 2021-07-21
Make infoview always enabled.

### LINKED ISSUE #101: When disabled `infoview` module, throws `:lua` error after attempting to open
First noticed this because of the fact that [suggested mappings are set regardless of whether their corresponding modules are `enable = true`'d](https://github.com/Julian/lean.nvim/blob/afdd047db65da46780825e4b303b3dbabbebf6bf/lua/lean/init.lua#L15).

For example, if you set `enable = false` for the `infoview` module, if you then try `<LocalLeader>i` in a Lean file you'll get a `:lua` error.

~~I believe we should fix this on a fundamental level by disabling calls to the public functions of a module if it has been disabled - using metatables, I think this should be doable without having to change the functions themselves.~~
EDIT: see discussion below

rish987:
Closes #101.

### DIFF
--- a/lua/tests/infoview/run_spec.lua
+++ b/lua/tests/infoview/run_spec.lua
@@ -3,7 +3,7 @@ local get_num_wins = require('tests.helpers').get_num_wins
 local fixtures = require('tests.fixtures')
 
 require('tests.helpers').setup {
-  infoview = { enable = true },
+  infoview = { autoopen = true },
 }
 describe('infoview', function()
   vim.api.nvim_command('edit ' .. fixtures.lean3_project.some_existing_file)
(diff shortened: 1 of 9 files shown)


Julian:
Awesome thanks!

Maybe we should add the test for the bug? I.e. where autoopen = false, making sure opening the infoview works without errors?

lua/tests/infoview/run_spec.lua
```diff
@@ -189,73 +189,16 @@ describe('infoview', function()
...
-    it('no auto-open for irrelevant file',
-    function(_)
-      local num_wins = get_num_wins()
-      vim.api.nvim_command("tabedit temp")
-      assert.is.equal(num_wins + 1, get_num_wins())
-      assert.is.equal(1, #vim.api.nvim_tabpage_list_wins(0))
-    end)
   end)
 
+  -- TODO
```
Julian, line 194: Is this still a TODO?

Julian (approved):
Looks great, thanks!

rish987:
Thanks -- I'm actually still working on a more general refactor of the tests (and also regarding that TODO will put that in its own spec). Will update soon!

rish987:
(actually, will make a separate PR/issue for this)

PR #103, opened 2021-07-23
Infoview autoclose

### LINKED ISSUE #43: Configurably autoclose the infoview if it is the last remaining window
(Moving this to its own issue from [this comment](https://github.com/Julian/lean.nvim/issues/20#issuecomment-847810221))

Add a configuration option `infoview.autoclose = true (default)` which indicates the infoview should be closed if it is the last remaining window.

Possibly a second consideration, once #35 is merged and there are multiple infoviews, is whether there should be an option that autocloses infoview windows when their Lean source file is closed (i.e. even if they are not the last remaining window).

rish987:
Working on this made me realize that we'll probably want to eventually decouple the concept of an infoview (which is just contains the pure tactic/term state data) from the "infoview view" window (the thing that can be opened/closed). Will work on this once we actually have some kind of persistent state feature with the infoview (e.g. #84).

Closes #43.


lua/lean/init.lua
```diff
@@ -70,8 +70,8 @@ function lean.use_suggested_mappings(buffer_local)
 end
 
 --- Is the current buffer a lean buffer?
-function lean.is_lean_buffer()
-  local filetype = vim.opt.filetype:get()
+function lean.is_lean_buffer(buffer)
+  local filetype = vim.api.nvim_buf_get_option(buffer or 0, "ft")
```
rish987, line 74: I'm a big fan of the OOP options but unfortunately I couldn't find a way to specify a particular buffer when using them.

rish987:
Also while working on this I've already started to see intermittent MacOS failures (e.g. [this one](https://github.com/rish987/lean.nvim/actions/runs/1059829721)), so we may want to ask a maintainer about nvim-lua/plenary.nvim#186 soon!

Julian:
Honestly I don't see much purpose in having two concepts personally.

We have vim concepts already of windows and buffers. An infoview is a lean concept, if it has some contents, that can be a buffer to me.

rish987:
Will create an issue for this, but how this is useful will become more clear once we've implemented #84. The idea is to be able to swap between infos with different pins (using a telescope picker). Figured we might as well decouple immediately (#105) because it makes the code easier to understand.

Julian:
It's likely easier to start from scratch here given the conflicts, so closing this one.

Issue #94, opened 2021-07-12
Option to run tests sequentially (and as the default in CI)


rish987:
I've [observed MacOS CI failures](https://github.com/rish987/lean.nvim/runs/2890467750?check_suite_focus=true) in the past that were fixed by forcing tests to run sequentially. Fundamentally I believe the issue is that we're trying to run multiple Lean 4 servers for the same workspace at once. If we get to specifically wanting to test this (which seems like an unlikely scenario anyways), it  should be its own unit test.

Sequential tests may be a (little) bit slower, but I'd say it's worth it to make things more deterministic!

Pending nvim-lua/plenary.nvim#185.

Julian:
Yeah... I've seen this too :/ Thanks for taking it upstream, I guess it's a reasonable option at least until we hit #93.

rish987:
At this point I can't run tests locally without things timing out 😭 

Probably because my computer's on the slower side. Shouldn't be an issue for me once this is resolved though. Anyways I prefer the test results being printed in a deterministic order...

Julian:
They're not speedy on mine either...

If you increase the timeouts they pass?

rish987:
For me, no. Even if I increase the limit on `wait_for_ready_lsp()`, `wait_for_line_diagnostics()`, etc. it hits plenary-busted's built-in 50 second limit when attempting `infoview/lsp_spec.lua`.

Issue #93, opened 2021-07-11
Make unit tests independent from one another (unordered)


rish987:
Generally, the unit tests in a particular `*_spec.lua` file must be run in the particular order given in the file, as subsequent tests may use the state built by previous ones (see esp. `infoview/run_spec.lua`). While I would still consider them unit tests, this makes them a bit less convenient to work with because it forces you to fix failures in a particular order (as failures after the first one may just be a side-effect of previous failures).

If we were to try to handle this at present (without changes to `plenary.nvim`), it seems we would have to either:
1. tear down `nvim` state after every unit test, or
2. put unit tests into separate files so they're run in separate instances,

both of which seem rather unwieldy.

So the best way to go about this seems to be to modify `plenary.nvim` to somehow allow for multiple nvim instances in a single file. This would also allow us to combine our `infoview/*_spec.lua` files into a single file.

Let me know if you have any other ideas!

Julian:
Yeah I briefly was thinking about this too but haven't spent much time trying anything specifically.

The third option (or 2.5th option) is having a bunch of fixtures which set up the state that each test is looking for, but which do so without necessarily saying whether they reuse or redo some state.

Bit hard to describe in text, but in Python the library that does this is e.g. https://pypi.org/project/testresources/ (or https://pypi.org/project/fixtures/ which is roughly the same library).

We could I guess write a simple version for neovim+Lua (and possibly coordinate with @tjdevries or the broader community since it'd be useful beyond just here).

Julian:
Closing this for now as I've done the best I can with some commits -- there's no other low-hanging fruit here that I can see, so we just wait till tooling upstream matures I think rather than futzing too much additionally here.

PR #104, opened 2021-07-24
Declutter infoview tests.

rish987:
Should also make testing infoview-specific stuff much more deliberate and complete!


lua/lean/infoview.lua
```diff
@@ -184,9 +184,11 @@ function Infoview:is_empty()
 end
 
 --- Close all open infoviews (across all tabs).
-function infoview.close_all()
+function infoview.close_all(pre_close_hook, post_close_hook)
```
Julian, line 187: Is there really a use case for this? It seems to me to complicate a rather simple function.

lua/tests/helpers.lua
```diff
@@ -116,11 +122,130 @@ local function has_all(_, arguments)
...
+    ("prev_win: %s, win: %s, win valid: %s"):format(vim.inspect(prev_win), vim.inspect(win),
+      vim.inspect(win and vim.api.nvim_win_is_valid(win))),
+    "is_open: " .. vim.inspect(this_infoview.is_open),
+    ("num_wins: %d, last_num_wins: %d"):format(helpers.get_num_wins(), last_num_wins)
+  }, "\n")
+
+  return result
+end
+
+local function open_infoview(state, arguments)
```
Julian, line 155: Can we try to avoid passing boolean flags to functions? The tests below are quite hard to read because it's hard to know what `assert.open_infoview(false, true)` even means.

lua/tests/helpers.lua
```diff
@@ -116,11 +122,130 @@ local function has_all(_, arguments)
   return true
 end
 
+local prev_buf_max = -1
```
Julian, line 125: This seems quite fragile to me because it adds global mutable state to all infoview-touching test runs. Would really prefer if we avoided doing that, we're already struggling with #94 showing us we have too much mutable state.

lua/tests/infoview/run_spec.lua
```diff
@@ -83,128 +71,102 @@ describe('infoview', function()
...
       infoview.get_current_infoview():close()
-      vim.api.nvim_set_current_win(new_win)
-      local num_wins = get_num_wins()
-      new_infoview_info = infoview.get_current_infoview():open()
-      assert.is_false(vim.api.nvim_win_is_valid(infoview_info.window))
-      assert.is_true(vim.api.nvim_win_is_valid(new_infoview_info.window))
-      assert.is.equal(num_wins + 1, get_num_wins())
-      assert.open_infoview()
+      assert.closed_infoview()
+      vim.api.nvim_command("tabnext")
```
rish987, line 135: TODO: would be good form to do `assert.changed_win()` here (and below), but it's actually not an issue in this specific case because we aren't actually closing anything.

rish987:
@Julian Thanks for the nudge. What I had before was actually a pretty bad take, sorry for making you read through it. I pretty much overhauled this in the recent commit, it should hopefully be much easier to follow now! And at this point I think it gives 100% coverage of what I wanted to test.

lua/tests/helpers.lua
rish987: This is something that you'd "opt-in" to on certain tests, and should be backwards compatible with the way we did things before (see e.g. `infoview/lsp_spec.lua`). ISTM that mutable state is pretty much the only way to get this kind of coverage, and we did have mutable state before, it was just littered throughout the tests themselves rather than centralized in a helper.

Did you perhaps mean #93? I'm pretty sure at this point the best way to handle it would be to:
1. Split up `infoview/run_spec.lua`.
2. Add a feature to `nvim-lua/plenary.nvim` that allows tests to early-abort as soon as an `it()` fails, so only the first error is shown.

rish987:
Will add more assert messages and split up `infoview/run_spec.lua` tommorrow!

rish987:
Okay, should be good for a review now! The only last thing I want to do is add a few helpful assertion failure messages, will do soon. Also regarding #94, see aa8b0db, nvim-lua/plenary.nvim#196 and nvim-lua/plenary.nvim#197.

Julian:
Nice! Thanks for more fun upstream work, well done. Yeah sounds good to merge here then to me!

PR #105, opened 2021-07-27
Infoview refactor: decouple "info" from "info view" 

rish987:
In support of #84 and per-window infoviews.

Let me know if you can think of a better name than "InfoviewView" haha....


rish987:
Thoughts on just calling the view/window "Infoview" and the data/buffer "Info"? 🤣

rish987:
Yeah, decided that Info/Infoview was far better than Infoview/InfoviewView...

Also plan on implementing buffer tracking for testing, similar to the window tracking we have right now, and eventually upstream both of these to `nvim-lua/plenary.nvim`.

lua/lean/infoview.lua
```diff
@@ -22,25 +27,31 @@ local _DEFAULT_WIN_OPTIONS = {
...
+
+--- A "view" on an info (i.e. window).
 local Infoview = {}
+local next_infoview_id = 0
 
---- Get the ID of the infoview corresponding to the current window.
-local function get_id()
-  return vim.api.nvim_win_get_tabpage(0)
+--- Get the ID of the current infoview.
+function infoview.get_id()
```
Julian, line 39: I'd very much rather this wasn't public. It seems like it's possibly just public for some testing reason, which I haven't fully understood yet.

rish987:
This should be the last "structural" change to the test helpers!

rish987:
Whoops, just noticed that `pre_closed_hook` stuff crept back in here... removed.

lua/tests/infoview/lsp_spec.lua
```diff
@@ -3,22 +3,22 @@ local helpers = require('tests.helpers')
 local fixtures = require('tests.fixtures')
```
rish987, line 4: Note: I'll soon update this file to properly make use of our new testing code.

lua/lean/infoview.lua
```diff
@@ -224,8 +243,10 @@ end
 
 --- Open an infoview for the current buffer if it isn't already open.
 function infoview.maybe_autoopen()
-  local id = get_id()
-  if not infoview._by_id[id] then infoview._by_id[id] = Infoview:new(id, options.width, options.autoopen) end
+  local tabpage = vim.api.nvim_win_get_tabpage(0)
```
Julian, line 246: Why move towards storing things by tabpage when there will be multiple views on one tabpage?

lua/lean/infoview.lua
```diff
@@ -22,25 +29,36 @@ local _DEFAULT_WIN_OPTIONS = {
...
 end
 
 --- Get the infoview corresponding to the current window.
 function infoview.get_current_infoview()
   return infoview._by_id[get_id()]
 end
 
+--- Get the info corresponding to the current window.
+function infoview.get_current_info()
+  return infoview._info_by_id[infoview.get_current_infoview().info]
```
Julian, line 52: Should just be able to make `Infoview.info` be the info and then this function isn't needed if I follow.

lua/lean/infoview.lua
```diff
@@ -118,8 +128,24 @@ function Infoview:focus_on_current_buffer()
   end
 end
 
---- Update this infoview's contents given the current position.
-function Infoview:update()
+function Info:new()
+  local new_info = {id = next_info_id, bufnr = vim.api.nvim_create_buf(false, true)}
+  infoview._info_by_id[next_info_id] = new_info
```
Julian, line 133: It seems we never remove anything from `_info_by_id`, so if that's the case, it's just a growing array-like table and you may as well just get rid of `next_info_id` and use `table.insert` + `#_info_by_id`, or possibly just `vim.tbl_add_reverse_lookup` and have all the indexes and infos in one table.

Julian (commented):
I still don't really understand what this is trying to do, and I don't personally really have a use for #84, so my review comments are fairly superficial.

Feel free to merge I guess though with or without them, given I'm not sure I can give much time on this one.

lua/lean/infoview.lua
rish987: Assume you're referring to per-window infoviews? This is where we'd initialize separate infoviews for both the tab and the window.

lua/tests/infoview/autoopen_spec.lua
```diff
@@ -33,7 +33,7 @@ describe('infoview', function()
     assert.buf.created.tracked()
     assert.win.created.tracked()
     vim.api.nvim_command("edit " .. fixtures.lean3_project.some_existing_file)
-    assert.buf.left.created({infoview.get_current_info().bufnr}).tracked()
+    assert.buf.left.created({infoview.get_current_infoview().info.bufnr}).tracked()
     assert.no_buf_track.initopened.infoview()
```
rish987: Yeah, this `no_track_buf` stuff is a bit less than ideal, I shouldn't have had to come back to modify this. I think a solution would be something like `assert.buf.left.tracked_pending()` to store an assertion state that is merged in on the next `assert. ... .tracked()` run. Will do this soon!

rish987:
No problem @Julian! The immediate point of this is just to simplify the code (from my perspective at least). Tried to address your comments, but yeah if those changes look good to you I'll go ahead and merge.

rish987:
P.S. this is also probably the single biggest PR I'll make to this repo, everything after this should be much more manageable!

lua/lean/infoview.lua
Julian: `get_id` semantically before meant, to an internal maintainer (i.e. us), "this is the index to store or retrieve an infoview from". Now this function reaches across that boundary and touches arbitrary state (which other functions rely on), so now that implementation detail is scattered again in multiple places, when I was hoping we'd go in the other direction and eventually remove `get_id` entirely. But it's fine, for now we need to get past continually refactoring this file without functional changes -- let's get to where there's functionality being changed and then I'll worry about it.

Julian:
Thanks, let's merge, appreciated.

PR #106, opened 2021-07-28
Tests: add buffer tracking.

rish987:
This was surprisingly easy to implement! Will upstream to `nvim-lua/plenary.nvim` after we've used it for a bit and I've added some helpful assertion messages. Should be easy to track tabs as well if the need arises.


Julian (approved):
lgtm

PR #107, opened 2021-07-29
Tests: refactor buffer/window tracking to use modifiers.

rish987:
Cleans things up a bit; also better enforces keeping buffer/window changes updated.


Julian (approved):
This is fine too I suppose.

Issue #92, opened 2021-07-09
Infoview updates slower in the presence of matchup (or any timer), and not at all when autocomplete pum is open


Julian:
We're running into two quite annoying neovim bugs.

One is where if you have e.g. [`vim-matchup`](https://github.com/Julian/dotfiles/blob/main/.config/nvim/init.vim#L50) enabled (as I personally do), or any other plugin which sets a [timer](https://github.com/neovim/neovim/blob/c1120ad0e1adf8b34ee10b63b2134fcb0d580fce/runtime/doc/usr_41.txt#L980-L985), then updates via `updatetime` see degraded performance (i.e. "the infoview updates slower").

The upstream neovim bug is [this one](https://github.com/neovim/neovim/issues/12587), and there are various downstream tracking bugs e.g. in matchup. There's a proposed [hack to fix via FixCursorHold.nvim](https://github.com/antoinemadec/FixCursorHold.nvim), but we'd need to tweak things a bit to recommend that (mostly because right now we use `setlocal updatetime` to configure the `CursorHold` time, whereas to use `FixCursorHold`, we would need to set a global option, `cursorhold_updatetime`). It may be that that's indeed worth doing though.

The second is even more annoying, which is that with `compe` enabled for autocompletion, `CursorHoldI` seems to not fire, or fire very sporadically, which means that in insert mode after calling `require'compe'.setup{}`, the infoview updates super unreliably.

I have to minimize what's needed to observe that behavior (it's not present in the minimal vimrc from this repo which is used for test cases), but I've already narrowed down that it is indeed related to enabling compe (which the tests do not do).

gebner:
> but I've already narrowed down that it is indeed related to enabling compe

FWIW, I've enabled compe and the infoview updates just fine for me.
```lua
require'compe'.setup {
  enabled = true;
  autocomplete = false;
  debug = false;
  min_length = 1;
  preselect = 'enable';
  throttle_time = 80;
  source_timeout = 200;
  incomplete_delay = 400;
  max_abbr_width = 100;
  max_kind_width = 100;
  max_menu_width = 100;
  documentation = true;

  source = {
    path = true;
    buffer = true;
    calc = true;
    nvim_lsp = true;
    -- emoji = true;
  };
}
```

Julian:
Thanks, I'll compare my settings to yours.

Julian:
[...]

PR #113, opened 2021-08-10
Pinnable/pausable infoview messages and Lean 4 infoview `processing...` message

### LINKED ISSUE #84: Support pinnable infoview messages

### LINKED ISSUE #85: Lean 4: show `Loading...` message when language server still processing file
(probably wise to implement this after #84)

Like in VSCode, we should maintain a per-URI state describing the point in the file that the language server has reached, which is [initialized to zero](https://github.com/leanprover-community/vscode-lean4/blob/014085d2b50e2fd1bd1bac78b94f3b0809dc08aa/src/leanclient.ts#L159) for each URI, [updated on `fileProgress` server notification](https://github.com/leanprover-community/vscode-lean4/blob/014085d2b50e2fd1bd1bac78b94f3b0809dc08aa/src/leanclient.ts#L176), and [cleared once the file has been parsed](https://github.com/leanprover-community/vscode-lean4/blob/014085d2b50e2fd1bd1bac78b94f3b0809dc08aa/src/leanclient.ts#L174) (or some other indicator).

We can then short-circuit attempting to update the infoview if the current URI hasn't been fully processed yet, printing a `Loading...` message, and having the URI progress state trigger an update for all relevant messages on parse completion for that file.

rish987:
Yeah this one's a bit on the bigger side as well, but I just couldn't justify shipping any one of these without the other. Should put us just a few lines away from #84. Feel free to take your time with this.

Closes #84, #85 #92. ~~Pending neovim/neovim#15276.~~


doc/lean.txt
```diff
@@ -60,39 +60,6 @@ health.check()                                                *health.check()*
 
 
 
-================================================================================
```
rish987, line 63: Not sure why this happened.

lua/lean/_util.lua
```diff
@@ -50,4 +50,48 @@ function M.subprocess_check_output(opts, timeout)
...
+    else
+      -- within modified range, invalidate
+      new_pos = nil
+      return new_pos
+    end
+    ::next_change::
+  end
+
+  return new_pos
+end
```
rish987, line 95: @gebner The logic for updating pin positions in VSCode [seemed to be lacking a bit](https://github.com/leanprover-community/vscode-lean4/blob/b7382e3b684baa2c211cf8f4d8e7f1681f99311c/lean4-infoview/src/infoview/infos.tsx#L40), this seemed much more accurate, but let me know if there's a particular reason it was done that way which I'm overlooking.

rish987:
(ready for review but marking as draft since we're waiting on neovim/neovim#15276; in general from now on in this repo I'll prefix titles with [WIP] in addition to draft if it's actually not ready for review)

lua/lean/_util.lua
gebner: The reason why the vscode extension parses the LSP notification is because Wojciech wanted to make the infoview as editor-independent as possible and vscode doesn't have an API to track document positions after changes.  (So you'd either parse the changes from the LSP event or from the vscode event.)

Is there a reason we can't use extmarks in neovim?

rish987: Yep we're doing the same here, the parameters used by this method come directly from `textDocument/didChange`. I thought this would be preferable to something built-in because it guarantees that pin positions are kept exactly in sync with the server's version of the text.

In particular if we are debouncing changes (have yet to write tests for this) if the pin positions are updated before sending `textDocument/didChange` and `$/lean/fileProgress` triggers a request, it could use positions that are invalid from the server's perspective.

Having said that, extmarks are new to me so yes I may have reinvented a bit here, will look into whether we can use this.

gebner: > if the pin positions are updated before sending textDocument/didChange

This is indeed a problem that is avoided by tracking the LSP change notifications.  But I'm not sure how much we gain by reimplementing extmarks just for pins alone.  All other commands (hover, autocomplete, etc.) that we provide assume that the document is in sync.

rish987:
Notice that we've (not so) incidentally completely closed #92 at 9d9a258 because we're updating on `textDocument/didChange` (as we should):
[asciicast](https://asciinema.org/a/LDWWS1nlqaXagHW5xIj32Dia1)
this is muy muy bueno

Julian:
Nice! Congrats!

lua/lean/_util.lua
rish987: Done (removed the custom math here in favor of extmarks). But I still think we should have some extra assurance about the document being in sync for updating the infoview (which we do have right now), since we'll want that to be able to happen at any time (e.g. on `$/lean/fileProgress`). So it's a bit different from the other commands you mentioned , since they're more directly interactive and so you have more of a guarantee of things being in sync (particularly if they happen after a timer).

rish987:
Alright this should be in a fairly stable state at this point, got to all the remaining bits I wanted to address. Let me know if there's anything else you want me to get to before this goes out.

The good news is that this finally makes this plugin something I'll be seriously using, so you can expect more Lean-specific PRs from me very soon!

rish987:
Okay, just removed the dependence on neovim/neovim#15276, it stands to reason that with #114 and the delay included on all update requests we don't need to worry about making a request before `textDocument/didChange`. There's still the issue of #115 but I have an idea of how to fix it (by simply using a timer delay of the debounce timeout on our request if we find pending changes).

More good news is that you can easily try this out for yourself now!!! With just a simple checkout, no need to change your nvim branch. Will test it myself over the next few days by digging into `mathlib`/`mathlib4`, but if you've tested it and have looked over the code yourself feel free to merge it.

rish987:
Merging for now, feel free to fix/improve things as you notice them.

PR #114, opened 2021-08-12
Infoview pin decouple and replace `CursorHold(I)` with `CursorMoved(I)`

### LINKED ISSUE #78: Outdated infoview updates can override the most recent one.

rish987:
Decouple refactor separated from #113. I think `CursorMoved(I)` better reflects our intentions than `CursorHold(I)` -- used a timer to approximate (improve upon?) "`delayedThrottled`" in VSCode. Closes #78 (actually this time) and might help with the first half of #92? Don't have the setup for this so @Julian let me know if it does.


Julian:
Thanks. I experimented a bit with this kind of change myself, so I suspect it seems reasonable to me, but I don't follow what it has to do with pins?

rish987:
Yeah no relation to pins, I'm just used to the new code and wanted to get your comments on that as well. Can make a new PR if you'd like.

Julian:
I mean I still don't really understand where we're going with these, we're getting what feels like quite-non-vim-like looking objects.

I had a local refactor which removes some of the ID juggling (on Infoviews and Infos, which I still maintain are "windows" and "buffers") but I still can't tell what Pins really are going to be, they seem like they may just want to be splits, but it's hard to tell.

But you seem to have a very specific future in mind so I'm still kind of just waiting to see what you're proposing, it's just hard to do so when things are tied up with other functionality -- here I don't know what part of this fixes #78, or whether the `CursorMoved` change is aesthetic or actually improves user-facing behavior (I couldn't find a case where it mattered when I tried it) -- but I think #78 is an important thing to fix, so if this fixes that, do feel free to merge, but yeah it'd be really really helpful to get to some functional changes.

Julian:
Sorry my split comment is probably obtuse so will elaborate on that bit -- I see pins basically mean "be able to keep some infoview state visible even when you keep moving around" -- in vim, that's "create a split" for normal things (usually for a persistent buffer), but here similarly for a transient one. So it'd seem to me an implementation is to be able to "orphan" an infoview and move to a new one, i.e. `.pin()` just splits, keeps the top infoview non-updating now, and starts to update the new split, without any new conceptual thing. But yeah maybe I'm still missing something, as I say, to me this isn't functionality that's top of mind, I have other things that bother me more in other areas, so I'm quite willing to go with your proposals if you're confident and use this yourself, I just can't say I follow the implementation yet.

Julian:
> Really I'm just trying to copy-cat VSCode's implementation with a few improvements

I figured this was part of the picture -- I explicitly *don't* want to do this -- I use vim because I like vim, not VSCode. Certainly taking the best parts of the Lean VSCode extension is a great idea and will help, but just because something works one way in VSCode isn't necessarily a reason it should work the same here I think.

> This is getting philosophical, but I could see infoviews/infos/pins being useful beyond Lean, for example if you want to keep some hover data visible without having to go back to a particular location.

In the current implementation, I honestly don't see the point. Vim has good ways of navigating windows. You lose all that if you just have text sitting in a buffer mashed together. What's wrong with a window?

rish987:
Yes I think we should allow for anything and everything! I've kept things very general so it's just a matter of providing different variants of the render function.

Issue #23, opened 2021-05-25
Make goto definition and hovering and other LSP commands work in the infoview


Julian:
In the infoview, it'd be nice if `vim.lsp.buf.definition()` (or hovering, etc.) on a term which is defined in the corresponding buffer functioned similarly to doing so in the original buffer -- i.e. jumps to the definition, etc.

Doing this for terms embedded in e.g. `try this:` messages would also be nice albeit require a bit more work.

gebner:
While this would indeed be nice, there is zero server support for this.

In Lean 3, we had widgets.  These are HTML fragments sent from the server (implemented using a wonderful React-like architecture).  The default HTML renderer for goals then included an event listener for clicks.  When clicked a request went back to the widget in the server, and then the widget told the editor to navigate to the location of the definition.  This was only ever implemented in vscode, as it is built on HTML and React.

For Lean 4, Sebastian has suggested (in private communication) to use virtual files for goals (for better portability between editors).  The server would then say for every goal: this is `lean://12345.lean` with file contents `.... ⊢ ...`.  You could then use standard LSP requests on these virtual files (such as hover or go-to-definition).  We would also like to have an LSP request to e.g. show implicit arguments, maybe by rewriting the virtual file, or by creating a new virtual file where the implicit arguments are expanded.

Julian:
Thanks as usual for the context!

In the interim, doesn't cheating work similarly to what it sounds like the server will ultimately do? I.e. on the client side, maintain a hidden buffer where we copy the contents of the original Lean buffer, and then in the infoview buffer is interacted with, delegate to running something on that hidden buffer? The issue would be making it look like LSP commands but even if we just directly mapped `gd` in the infoview that may already help.

gebner:
> doesn't cheating work similarly to what it sounds like the server will ultimately do?

This is harder than it sounds.  If you maintain a hidden buffer with contents of the original lean file, then the server will parse it again (which takes time and cpu).  If the hidden buffer is empty, you need to replicate at least the `import` and `open` statements.  Then there's the question of how to copy the infoview contents to the buffer.  The goal state `... ⊢ ...` won't parse as it is.  It might parse some of the time if you liberally sprinkle `∀` over it, but it's still super flaky.

I think this would be a lot of work for an unsatisfactory implementation, and I definitely wouldn't prioritize it.

> The issue would be making it look like LSP commands but even if we just directly mapped gd in the infoview that may already help.

[...]

PR #118, opened 2021-08-16
Support for RPC protocol

gebner:
This is a prototype implementation of the RPC protocol, which uses the `getInteractiveTermGoal` function to obtain the term goal from the server and renders it as a string.

~~The protocol is still subject to change, so don't merge this yet.~~

- The docgen tool seems to have a problem with the `Session:call` function, not sure what's going on there.
- ~~We now have all the tools necessary to e.g. show implicit arguments, docstrings, types, etc. in the infoview (from the InfoWithCtx object).  Is there any obvious way to implement these interactive actions in neovim?~~
- ~~The vscode version uses a `FinalizationRegistration` to hook into the javascript GC to detect when remote references can be released.  I've tried adding a `__gc` metatable, but this doesn't seem to have an effect in neovim (even though it works just fine on the command-line lua).  So we just store all references in the subsession object, and close the subsession whenever the pin is moved (which releases the references).~~
- ~~The code still uses `buf_request` with a bufnr (hardcoded to 0), which will break if there are multiple LSP servers.  I should figure out how to get the LSP client object and send requests directly (and only) to the lean server.~~


Julian:
> We now have all the tools necessary to e.g. show implicit arguments, docstrings, types, etc. in the infoview (from the InfoWithCtx object). Is there any obvious way to implement these interactive actions in neovim?

I am playing with some things here as part of toying with widgets. I don't have concrete ideas yet but maybe they're the same UX problem. Will play around with this branch. Anything special I need other than a recent Lean 4?

> The docgen tool seems to have a problem with the Session:call function, not sure what's going on there.

It's cranky occasionally (e.g. I can't get it running locally) :/. But will see if I can tell what's up.

> figure out how to get the LSP client object

`vim.lsp.buf_get_clients()` I think? There's possibly multiple though if someone attaches e.g. our client as well as say a linter one, so you have to postfilter I think.

gebner:
> I am playing with some things here as part of toying with widgets. I don't have concrete ideas yet but maybe they're the same UX problem.

I think the UX problem is the same.  If you have a nice way to render Lean 3 widgets and handle keypresses, then the same idea should also work here.

> Anything special I need other than a recent Lean 4?

You need this branch here, none of this upstream yet: https://github.com/leanprover/lean4/pull/596

gebner:
See also #23.

rish987:
I'm going to be trying this out as well, and will be happy to fix any conflicts related to #113.

> I've tried adding a __gc metatable, but this doesn't seem to have an effect in neovim (even though it works just fine on the command-line lua)

This could be a Lua 5.1 vs. Lua 5.2 issue, have you tried [this workaround](https://stackoverflow.com/questions/27426704/lua-5-1-workaround-for-gc-metamethod-for-tables)? Again I'll be looking into this myself and so may give it a shot soon.

gebner:
> This could be a Lua 5.1 vs. Lua 5.2 issue (LuaJIT is frozen at 5.1),

I think you're right!  I didn't realize that there are version differences in Lua.

lua/lean/rpc.lua
```diff
@@ -86,7 +87,8 @@ local function connect(bufnr)
 end
 vim.lsp.handlers['$/lean/rpc/connected'] = function(err, _, params, _, _, _)
   if err ~= nil then return end
-  local bufnr = vim.uri_to_bufnr(params.uri)
```
rish987: This is not a big deal, it's just that `uri_to_bufnr` looks like it can create a new buffer if one doesn't already exist for that URI, which might be unexpected.

rish987:
Got `__gc` working with that `userdata` workaround. I tried it out for a bit and the refs seem to be getting released at a reasonable rate (definitely not immediately after moving the pin), but of course I'll leave this to your judgment, feel free to add some `collectgarbage()` calls if you think it will help.

Going to work for a bit on simulating some kind of HTML `divs` for our Info to use in the `Info:render()` function so that we can easily map from cursor position in the Info to the relevant objects. I think this should actually be pretty simple to do! Then we should be able to just use keymaps and hover windows in the infoview to get the extra information (though I'm not yet familiar with the syntax for using these refs, so will leave that to you).

rish987:
Alright @gebner everything seems to be ready  for you at this point (ignore broken syntax highlighting):
[asciicast](https://asciinema.org/a/RqXHLLalM65Sd42HzzJQIGuWL)
I've mapped widgets to `<localleader><space>` for now, right now all that happens is it brings up a hover window showing the current div. The API returns the full div stack hover so you can traverse up to get more data, e.g. the `tag` and its ref in the case of `text` divs. Now what I think you'll want to do is `a.void()` wrap some code in `Info:widget()` to actually use the ref to do a particular request and display the result in a hover window. Again I don't yet know how to do this so definitely feel free to have a go at it.

Also @Julian I don't know much about Lean 3 widgets yet, but I believe this is a piece of the puzzle for that? Will work on separating the new `div` stuff from this PR so we can get Lean 3 widgets working ASAP.

lua/lean/rpc.lua
```diff
@@ -38,40 +46,81 @@ function Session:new(bufnr, uri)
...
+---@field textDocument TextDocumentIdentifier
+---@field position LspPosition
 
 ---@param pos TextDocumentPositionParams
 ---@param method string
 ---@return any result
 ---@return any error
 function Session:call(pos, method, params)
-  if self.session_id == nil then
+  if not self.connected then
```
rish987, line 107: Would there be any harm in making this a `while`?  Just in case it disconnects for some reason in one of the calls.

lua/lean/infoview.lua
```diff
@@ -154,7 +154,34 @@ function Info:widget()
...
+  local sess = nil
+  for _, div in ipairs(div_stack) do
+    sess = sess or div.tags.rpc_sess
+  end
+
+  for i = #div_stack, 1, -1 do
+    local info_with_ctx = div_stack[i].tags.info_with_ctx
+    if info_with_ctx ~= nil then
+      return a.void(function()
+        local info_popup = sess:infoToInteractive(info_with_ctx)
```
rish987, line 167: CMIIW but I think the div is the only thing with the reference to this, so it could get GC'ed immediately after making the popup. I think you may have not noticed this since it looks like these popups aren't interactive yet (will soon try this myself).

lua/lean/rpc.lua
```diff
@@ -6,7 +6,7 @@ local control = require'plenary.async.control'
 ---@class RpcRef
 
 ---@class Session
----@field bufnr number
+---@field client any vim.lsp.client object
```
rish987, line 9: Forgot to mention this earlier, but neovim [won't send a request unless the client supports it](https://github.com/neovim/neovim/blob/b35de6c525e7c8ad9d64d42f57f348dff608b9e5/runtime/lua/vim/lsp.lua#L1261), so since the requests we're making here seem to be specific enough to Lean 4, it seems to me that this commit and the next (419e8928cb78622480ef207d3bbc027e183e820b) aren't necessary. Did you run into an issue before this commit?

rish987 (commented):
Thanks! Just a few comments, wanted your thoughts on them but feel free not to act on them since anyways I'll soon be extracting bits from this into #123.

lua/lean/rpc.lua
gebner: As the commit message says:

> Workaround for leanprover/lean4#648

The server supports the notification, but crashes when you send it (after the file is closed).

gebner:
Closed in favor of #123.

PR #119, opened 2021-08-16
Use `lean --version` to infer filetype when `elan` is installed.

### LINKED ISSUE #88: Make standalone files use the Lean version from `elan default` when set / present

rish987:
Closes #88, ~~failed Lean 3 stdlib ft test pending leanprover/elan#36~~. Also fixed the stdlib filetype tests.

### DIFF
--- a/lua/lean/ft.lua
+++ b/lua/lean/ft.lua
@@ -1,5 +1,12 @@
-return {
-  detect = function()
-    vim.opt.filetype = require('lean.lean3').__detect() and 'lean3' or 'lean'
+local M = {}
+if vim.fn.executable"elan" then
+  M.detect =
+  vim.schedule_wrap(function()
+    vim.opt.filetype = require('lean.lean3').__detect_elan() and 'lean3' or 'lean'
+  end)
+else
+  M.detect = function()
+    vim.opt.filetype = require('lean.lean3').__detect_regex() and 'lean3' or 'lean'
   end
-}
+end
+return M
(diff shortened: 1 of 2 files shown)


rish987:
@gebner I think there might be a slight MacOS bug with leanprover/elan#36, see test failures -- since our default toolchain in CI is Lean 3 (and the error mistook `lean` for `lean3`), I have a suspicion that it has to do with leanprover/elan#36 checking the path case-sensitively, while MacOS is has a case-insensitive (all-lowercase) filesystem.

gebner:
> I have a suspicion that it has to do with leanprover/elan#36 checking the path case-sensitively, while MacOS has a case-insensitive (all-lowercase) filesystem.

This is a potential explanation.  But I think we need someone with access to a MacOS machine to debug this (and test the fix).

Julian:
I have a mountain of things on my plate now for the next few days so I'm going to be quite slow at many things, but will see if I can circle back when I get a moment.

rish987:
Nevermind Gabriel your code is good (also tested it at rish987/elan@e303f1d, let me know if that's worth upstreaming) I fixed this by `schedule_wrap`ing our `elan`- based filetype detection.

Julian:
Can you separate (and feel free to merge separately) the unrelated changes here so it's easier to see what this does?

PR #123, opened 2021-08-22
[WIP] Lean 3/4 Widgets

rish987:
Need to look a bit more into how this is done in VSCode. At the moment this just includes a refactor to render the info with `div`s, extracted from #118. I'll be basing our widgets on this, so let me know your thoughts on it.


lua/lean/infoview.lua
```diff
@@ -139,9 +174,8 @@ local paused_txt = "[PAUSED]"
 
 --- Update this info's physical contents.
 function Info:render()
+  self.div = html.Div:new()
```
Julian, line 177: I don't want sprinklings of HTML concepts throughout the codebase -- vim is not an HTML-based editor. We of course as part of this are translating from HTML to vim-native concepts as part of rendering widgets, but that should be limited just to the interface between widgets and the rest of the plugin (i.e. somewhere should be a component that renders HTML to vim APIs, but then nowhere else is there something HTML looking).

rish987: If you look at lean/html.lua, you'll find that this is just a way for mapping from infoview cursor location to the relevant stack of objects. This is useful beyond widgets for jump-to-location, copy-to-comment, etc.

Julian: Right -- we should have our own native way of doing so, and map the HTML we get from widgets into it, rather than vice versa (trying to map vim things into HTML even when we're not dealing with HTML handed to us).

rish987: Yep that's the plan, I don't intend to go any further than what we currently have at lean/html.lua.

Julian: What I mean is nothing in lean.infoview (or its components) should be importing or talking about HTML -- the widget handler should be the only thing dealing with HTML, and then passing either strings or more likely as you say some agnostic position-or-interaction-related objects around. `components.goal` should definitely not take an argument of type `html.Div` for instance.

Here in the PR you have lots of HTML objects already being passed around.

rish987: > passing either strings or more likely as you say some agnostic position-or-interaction-related objects around.

This seems to me like it would require creating new objects, which we should avoid if we can directly use what we already have.

> `components.goal` should definitely not take an argument of type `html.Div` for instance.

Yeah an argument isn't necessary there, but it should definitely at least _return_ an `html.Div`, otherwise we'd completely lose the context of separate goals.

> Here in the PR you have lots of HTML objects already being passed around.

I used it every where I could to preserve context, also I find it much cleaner than `table.insert`ing a `lines` array.

rish987: Perhaps from your perspective it's just a matter of renaming the `html` module to something else?

Julian: > This seems to me like it would require creating new objects, which we should avoid if we can directly use what we already have.

We don't already have anything -- either way we are writing a simple UI framework.

>  but it should definitely at least return an html.Div, otherwise we'd completely lose the context of separate goals.

Please lay out what you're trying to do -- I don't follow what this has to do with widgets (and there are again lots of seemingly unrelated things rolled up in the code that's here already, so I can't tell what direction you're trying to take this in.).

But no, I don't think it should return one either.

> Perhaps from your perspective it's just a matter of renaming the html module to something else?

Possibly -- one way or another we'll end up with some form of TUI, so yes the code I was playing with for widgets was using similar kinds of objects than one would find in HTML of course (a `Container` of elements, a `Button`, a `SelectableList`, etc.), so if that's what you mean by rename, sure.

rish987: > I don't follow what this has to do with widgets
Nothing to do with widgets, just wanted to fully integrate this as soon as we've introduced it.

> Please lay out what you're trying to do
One example is copy-to-comment

lua/lean/infoview.lua
```diff
@@ -117,14 +144,66 @@ function Info:new()
 
   vim.api.nvim_buf_set_name(new_info.bufnr, "lean://info/" .. new_info.id)
   vim.api.nvim_buf_set_option(new_info.bufnr, 'filetype', 'leaninfo')
+  set_augroup("LeanInfoWidget", string.format([[
+    autocmd CursorMoved <buffer=%d> lua require'lean.infoview'._info_by_id[%d]:__cursor_hold()
+    autocmd BufEnter <buffer=%d> lua require'lean.infoview'._info_by_id[%d]:__cursor_hold()
```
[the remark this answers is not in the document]
rish987: In normal mode enter means "go to the beginning of the next line", figured we shouldn't mess with that. But yeah will add a command to make this configurable.

lua/lean/infoview.lua
```diff
@@ -117,14 +144,66 @@ function Info:new()
...
+  if not div_stack then return end
+
+  local pin_div = html.util.get_parent_div(div_stack, "pin")
+  if not pin_div then return end
+  local pin = pin_div.tags.pin
+
+  lean3.widget(pin, div_stack, event)
+end
+
+function Info:__click()
```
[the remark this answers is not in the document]
rish987: See the new `:LeanInfoEvent <event_name>` command.

rish987:
@gebner got to all of your points (as well as my own), let me know what you think. Feeling pretty good about the code at this point. Will leave fixing documentation and testing to the end.

>  Storing the state is easy with floating windows: neovim stores the open floating windows and they're not removed when refreshing the infoview. But this is harder if we want to do inline popups

Good point. FWIW I think Lean 4's way of doing things with stateless widgets and RPC is far better than the Lean 3 way. I have an idea for inline tooltips w/ Lean 4 widgets, basically to be able to have a "filter" on a div whose return value replaces it when rendering it. Will work on integrating your changes from #118 into this PR, so hopefully you should soon be able to see what I mean.

rish987:
Lean 4 widgets seem to be working at this point, however I've noticed the RPC error `-32603`: `unknown free variable '_uniq.10'` when clicking on the notation for "or", for example. @gebner is this a known upstream issue?

TODO:
- [ ] release refs when pausing a pin (likely with an `html.Div` function to recursively clear all tags that we will call when pausing)
- [ ] I noticed that in #118 you made sure to clear the refs of a pin when closing its infoview -- in light of possibly multiple pins, I think we should take a more nuanced approach by letting pins keep track of whether they're visible, and automatically pausing when they aren't (so this ties into the point above), will likely put this off to after this PR
- [ ] currently Lean 4 tooltips aren't "resilient" to updates, that is on an update all tooltips will be cleared. So what I think we want to do is make use of the undo stack I've introduced here for Lean 3 in order to "replay" clicks on update (scratch the "filter" idea above, this seems like a more fundamental solution)
- [ ] new keymapping for the info that will show a pum from which you can select type, explicit expression, or docstring, and the tooltip will show only that one (right now click shows everything)

Julian:
I still have negative amounts of time at the minute :(

(I'm prepping for start of semester on a new job, so things are quite hectic...) -- please feel free to merge this as soon as you'd like (especially if @gebner is good with it), given I don't see myself being able to even look at it for at least a few weeks.

Julian:
Oh, also announce it on Zulip :)

rish987:
No problem! Yeah will do, there are a few more things to do here, but will likely be ready very soon, will let you know Gabriel when it is.

gebner:
I'm also on vacation.  Feel free to merge if it doesn't introduce any regressions.

> Lean 4 widgets seem to be working at this point, however I've noticed the RPC error -32603: unknown free variable '_uniq.10' when clicking on the notation for "or", for example.

cc @vtec234

Julian:
@rish987 I want to make sure you're not spinning wheels -- I think we need to try to avoid these long PRs.

But just to head off something, hopefully you're not planning on making widget support require or depend on hop.nvim?

Having it integrate with it is possibly fine if someone has it installed.

But yes please describe what you're trying to do? Widget support doesn't seem very related to easy motion plugins, albeit integrating with one optionally might be useful to some.

rish987:
Yeah sorry for the long silence here, I've just been coming accross bugs in the TUI code from time to time so didn't want to rush it. hop.nvim will be a soft dependency, with it you can very quickly jump to clickable divs, but you don't need to have/use it.

Though it works pretty well at the moment -- you can try it out, use "K" to click, "J" to enter a tooltip, "U" to undo (though there's a problem with undo in Lean 4 at the moment), "S" to hop (If you have phaazon/hop.nvim#153 installed).

Yeah sorry it's gotten so big, though pretty much all of the changes are in the new TUI code which I plan to upstream somewhere.

With the recent hop.nvim integration, I finally have all the features I want at this point! So just bug fixing and README documentation is left.

Julian:
Got it all that sounds great! And all good I'm the last one to complain, I haven't been able to touch Lean in a month :(

But especially the upstream part :D -- 100% agree that was exactly the idea I had as well, that we may as well have a TUI library others can use similar to some GUI frameworks that have popped up which only work in GUI-based neovims.

But yeah sounds great! No rush at all, and I do hope when I finally get a chance to get back to this to try it out, it definitely was very nice last time I gave it a shot.

gebner:
>  I think we need to try to avoid these long PRs.

I completely agree with Julian.  It would rather merge a minimal version of this PR with some remaining bugs in the widget code than have the PR lie around for another month.

Really the only wish I have before we merge this PR is that it doesn't introduce regressions:
 1. On really old versions of Lean 4 the infoview only shows "Processing" (this is before the fileProgress notification was introduced).
 2. When running neovim in a terminal (I don't see this in neovim-qt), there is an enormous amount of flickering in the infoview when moving around:
   - Move the cursor.
   - The whole expected type is colored gray and the `[LOADING]` message appears.
   - The `[LOADING]` message disappears and the colors come back.
 3. In general, I find the `[LOADING]` message to be very distracting.  (In vscode we had the same issue, so Ed added an easing animation to the progress color. The pin header takes a second to gradually change to orange.  I don't think we can or should do that in neovim, I just want to point out that this is a real UX issue.)

There also seems to be quite a bit of debugging output (e.g. the number of undo levels, or the "exprExplicit:" annotation I added for debugging in the popup).  This debugging output should either be removed or disabled (by default) with a configuration setting.

UI-wise I'm not a big fan of the new extra `-- PIN` header for the default pin.  It adds quite a bit of visual noise.  But we can bikeshed that after the PR is merged.

lua/lean/html.lua
```diff
@@ -0,0 +1,705 @@
...
+  return self:_div_from_path(path, setmetatable({}, {__mode = "v"}))
+end
+
+function Div:_div_from_pos(pos, stack)
+  table.insert(stack, self)
+
+  local text = self.text
+
+  -- base case
+  if pos <= #text then return nil, stack, {} end
```
gebner, line 163: When I click around randomly in the infoview (sorry, no better repro), I sometimes manage to get an exception from this line because `pos` can be nil here.

gebner: I am using neovim-qt but I can also reproduce it in the terminal (with `set mouse=a`).

And now I've figured out how to reproduce it: you just need to select the final `--` with the mouse.

rish987: I still can't reproduce it with the mouse. Are you still on 0.5? Perhaps an upstream issue is causing one of the failure states in `pos_to_raw_pos`?

rish987: Ah I see, it's when you're doing a visual selection with the mouse, right? I think I've reproduced it.

rish987: Yep, you can reproduce without the mouse as well when you do a visual selection and go past the end of any line.

Vtec234:
> Lean 4 widgets seem to be working at this point, however I've noticed the RPC error -32603: unknown free variable '_uniq.10' when clicking on the notation for "or", for example.

I can't reproduce this in the VSCode infoview so I installed this branch using `vim-plug`. But on opening any Lean file I get "Error scheduling vim.schedule lua callback: .../lean.nvim/lua/lean/ft.lua:4: attempt to index field 'opt' (a nil value)" . Any ideas?

Julian:
@Vtec234 sorry about that -- usually that means perhaps that you've called `lean.setup` twice, e.g. by resourcing your file? We want to fix that but haven't done so yet. If that's not it though let us know.

rish987:
@Vtec234 also feel fre to DM me on Zulip if you need help using working with this PR, I haven't documented much yet

lua/lean/html.lua
rish987: Should be fixed now.

rish987:
> On really old versions of Lean 4 the infoview only shows "Processing" (this is before the fileProgress notification was introduced).

What's a good nightly to test this on, and is there some handle on whether the server has this capability (perhaps on initialize response)? Also, I think this is a regression from #113 where I introduced this message, so will address in a separate PR.

gebner:
> What's a good nightly to test this on,

Pretty much any version before June.
mathlib4: 5b04a77d4578a75b0cfe3d7fc77502e1f334945f
lean4: 10cf2192139f5e29270ccbfff83af37b9b1dabb0
Nightly from 2021-06-03 should also work.

> is there some handle on whether the server has this capability (perhaps on initialize response)?

I don't think so.  The only way to tell is when you receive a fileProgress notification.  I think the easiest approach is just to change this line here: https://github.com/Julian/lean.nvim/pull/113/files#diff-b5bae483a95c014b5c285708801d28ec6aa7b0ee241a91736dd2467a7834f5d2R14
and assume that lean is finished if no fileProgress notification has been received yet.

rish987:
Right, so we could just return false there, but in newer versions of Lean 4, how will the server respond if we make a plainGoal/plainTermGoal/interactiveTermGoal request _after_ initialization but _before_ the first file progress notification? Is this an edge case we need to worry about? It may not be easy to reproduce if so.

gebner:
@rish987 I've pushed a commit to support interactive traces.  I get empty lines in the traces, is the div code inserting extra newlines?

> how will the server respond if we make a plainGoal/plainTermGoal/interactiveTermGoal request after initialization but before the first file progress notification?

It should respond with the ContentModified error.

lua/lean/infoview/components.lua
```diff
@@ -191,4 +203,72 @@ function components.diagnostics(bufnr, line)
...
+
+        div.divs = {}
+        div.highlightable = false
+        div.tags.event.click = nil
+        div:insert_new_div(tagged_text_msg_embed(expanded, sess))
+        return true
+      end
+      div.highlightable = true
+      div.tags.event = { click = click }
+      div:insert_div({}, '...')
```
rish987, line 248: I think we should have a corresponding `clear` event that sets it back to this state so that undo works here.

rish987:
> It should respond with the ContentModified error.
Gotcha, so this should be safe to do. Though note we're still waiting on neovim/neovim#15262.

> I get empty lines in the traces, is the div code inserting extra newlines?
The div code doesn't add any bytes that you didn't. How exactly can I reproduce this?

lua/lean/infoview/components.lua
rish987: Also, how can I get this collapsed trace to show so I can try it out?

gebner:
> I think we should have a corresponding clear event that sets it back to this state so that undo works here.

The vscode infoview I think adds an arrow `>` that changes to `v` when expanded (like a GUI treeview).  I didn't put too much effort into the UI, I just wanted to do the bare minimum to see full traces (the plain LSP diagnostics are truncated).

> Also, how can I get this collapsed trace to show so I can try it out?

```lean
set_option trace.Meta.synthInstance true in
example : DecidableEq (Option Nat) :=
  inferInstance
```

> How exactly can I reproduce this?

If you expand the trace above, you'll get an empty line.  (I didn't look very hard.  Maybe it is a Lean bug, maybe I'm holding the infoview API wrong, maybe it's something else.)

gebner:
I've pushed support for interactive goals.

rish987:
Awesome, thank you!!! I just want to get emmy annotations, docgen, and README updated and then this should be good to go.

rish987:
Update: I've got a situation that I need to deal with for the next few days to a week, so feel free to take over here and do the finishing touches and merge, or you can wait as well

Julian:
Let's merge.

Bikeshed away :)

(I'm hoping I get some free time this afternoon to play with this, but thanks again for all the effort on this one!)

Julian:
And hope everything is ok of course! Please take care.

PR #124, opened 2021-08-24
Add syntax highlighting for #reduce

AlexDikelsky:
I just started using lean.nvim, and lean in general. Could `#reduce` be highlighted the same way as `#eval` and `#check`?


Julian:
Cool! Thanks (and welcome) -- can you add this to `syntax/lean3.vim` as well (which covers Lean 3).

After that good to merge!

PR #129, opened 2021-09-02
Adapt LSP request handlers for neovim/neovim#15504.

rish987:


lua/lean/_util.lua
```diff
@@ -50,4 +50,33 @@ function M.subprocess_check_output(opts, timeout)
...
 end
 
+function M.wrap_handler(other_handler, handler)
+  return function(...)
+    other_handler(...)
+    handler(...)
+  end
+end
+
+function M.mk_handler(fn)
```
Julian, line 60: This looks like it's @mfussenegger's code from https://github.com/mfussenegger/nvim-lsp-compl/commit/29a81f3747373f42d002775b1f79dcd19fada2e4 right? If so it's GPL 3.0 so we at very least have to notate that here (unless perchance @mfussenegger has already said somewhere it's OK to crimp regardless?)

Julian (approved):
Small license issue but after clarifying that otherwise think this is great thanks for knocking it out!

lua/lean/_util.lua
rish987: Yep just copied it -- added acknowledgements, let me know it that is sufficient

mfussenegger: Fine for me to copy it.

Julian: Thank you (both)!

PR #135, opened 2021-09-21
Support neovim 0.5 again

gebner:
Restores 0.5 support.  The CI runs checks on both 0.5 and nightly now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,10 +57,4 @@ jobs:
           git clone --depth 1 https://github.com/nvim-lua/plenary.nvim packpath/plenary.nvim
 
       - name: Run tests
-        run: |
-          make test
-        if: runner.os == 'Linux'
-
-      - name: Run tests
         run: make test
-        if: runner.os == 'macOS'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,15 @@ jobs:
             echo "$HOME/.elan/bin/" >> $GITHUB_PATH
         if: runner.os == 'Linux'
 
+      - name: Install neovim
+        run: |
+          neovim_dir=$HOME/neovim
+          mkdir $neovim_dir
+          curl -sL https://github.com/neovim/neovim/releases/download/v0.5.0/nvim-linux64.tar.gz \
+            | tar xzf - --strip-components=1 -C $neovim_dir
+          echo $neovim_dir/bin >> $GITHUB_PATH
+        if: runner.os == 'Linux'
+
       - name: Install neovim and elan
         run: |
           brew update
@@ -49,8 +58,6 @@ jobs:
 
       - name: Run tests
         run: |
-          curl -OL https://raw.githubusercontent.com/norcalli/bot-ci/master/scripts/github-actions-setup.sh
-          source github-actions-setup.sh nightly-x64
           make test
         if: runner.os == 'Linux'
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,41 +9,31 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest]
+        nvim-version: [v0.5.0, nightly]
         lean-version:
           - "leanprover-community/lean:3.33.0"
     steps:
       - uses: actions/checkout@v2
-      - run: date +%F > todays-date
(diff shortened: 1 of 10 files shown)


gebner:
Not sure what's going on with the search path test.  The function only fails when called from the test, but not if open nvim and run it manually.

syntax/leaninfo.vim
```diff
@@ -16,7 +16,7 @@ hi def link leanInfoGoalHyp Type
 hi def link leanInfoGoalVDash Operator
 hi def link leanInfoExpectedType Special
 
-hi def link leanInfoError DiagnosticError
-hi def link leanInfoWarning DiagnosticWarning
-hi def link leanInfoInfo DiagnosticInfo
+hi def link leanInfoError LspDiagnosticsDefaultError
```
Julian, line 19: This'll break highlighting on the nightlies (see below for an example)

Screen Shot 2021-09-21 at 15 45 46

 -- these highlight groups don't exist anymore in them -- I think you have to do something like what I showed here: 

```suggestion
" Old neovims (0.5) pre https://github.com/neovim/neovim/pull/15585
if luaeval('vim.diagnostic == nil')
    hi def link DiagnosticError LspDiagnosticsDefaultError
    hi def link DiagnosticWarning LspDiagnosticsDefaultWarning
    hi def link DiagnosticInformation LspDiagnosticsDefaultInformation
endif

hi def link leanInfoError LspDiagnosticsDefaultError
```

to forward port the new ones to old neovim.

.github/workflows/ci.yml
```diff
@@ -27,6 +27,15 @@ jobs:
...
         if: runner.os == 'Linux'
 
+      - name: Install neovim
+        run: |
+          neovim_dir=$HOME/neovim
+          mkdir $neovim_dir
+          curl -sL https://github.com/neovim/neovim/releases/download/v0.5.0/nvim-linux64.tar.gz \
+            | tar xzf - --strip-components=1 -C $neovim_dir
+          echo $neovim_dir/bin >> $GITHUB_PATH
+        if: runner.os == 'Linux'
```
Julian, line 37: This makes the tests *only* run on 0.5 on Linux -- ideally we'd run both 0.5 and nightly (which is what that github-actions-setup.sh does) if we support both, which is the annoying bit I didn't want to deal with (modifying the matrix to parallelize that).

If you don't want to do the matrix modifying though at least perhaps leave both blocks so they run in serial.

syntax/leaninfo.vim
gebner: Then I was confused about these lines here: https://github.com/neovim/neovim/pull/15585/files#diff-8df5fbf5803bbea8d9e9d4d8a63f0bac3a32e1925d74990a43c6327fad2aad03R28-R48

I thought that would work for backwards compatibility.  I'll add the if.

.github/workflows/ci.yml
gebner: The https://raw.githubusercontent.com/norcalli/bot-ci/master/scripts/github-actions-setup.sh only downloads nightlies, so it won't help here.  But I'll add a matrix.

gebner: This was actually easier than I'd thought, but now we're downloading neovim release binaries on both ubuntu and macos (instead of using brew).  I hope that's ok for you.

syntax/leaninfo.vim
Julian: Interesting. That looks like you should be right, but yeah I don't get this behavior on a nightly. I'll see if I can tell why whilst also seeing if I understand the test failure you're seeing.

.github/workflows/ci.yml
Julian: Nice! That's easier than I thought it'd be too -- yeah that's great, I think the reason I didn't do it that way in the first place was because neovim was in the midst of moving over to GitHub actions a few months ago when I wrote this, so they weren't pushing nightlies, but this makes total sense now.

rish987:
(sorry, am working on generalizing phaazon/hop.nvim in support of #123, will look into test failures soon)

gebner:
Tests pass on 0.5 now, but the neovim nightly builds are gone temporarily so now the other tests fail :-( https://github.com/neovim/neovim/issues/15709

The only thing I've changed since yesterday is the backported `list_workspace_folders` function though, so the tests should pass with nightly once the builds are back up.

Julian:
Hooray, thanks!

PR #136, opened 2021-09-22
Fix publishDiagnostics handler override.

gebner:
I thought that the missing diagnostics came from the recent 0.6 support PR.  But it turns out that this is not the case, the diagnostics were broken for me because I've customized the `vim.lsp.handlers['textDocument/publishDiagnostics']` function, and the infoview code ignored that when overriding the handler.

The reason my earlier fix didn't work is because the diagnostics were still shown properly (using my overridden handler) when the file was first loaded.  Only subsequently added diagnostics were missing..


Julian (approved):
lgtm, thanks

PR #137, opened 2021-09-27
Test on neovim 0.5.1

gebner:


gebner:
Apparently the new diagnostics test is flaky...

PR #138, opened 2021-10-04
Tweak infoview widgets style.

gebner:
- [x] Less information by default (like pin number, undo list size, etc.)
 - [ ] Hover tooltip shows type instead of "click here"
 - [x] Map CR to click
 - [x] Clean up hover tooltip (no "exprExplicit:", etc.)
 - [x] Make highlight definitions work in GUI (only terminal colors provided atm)

### DIFF
--- a/doc/lean.txt
+++ b/doc/lean.txt
@@ -61,6 +61,11 @@ health.check()                                                *health.check()*
 
 
 ================================================================================
+infoview.enable_debug()                              *infoview.enable_debug()*
+    Enables printing of extra debugging information in the infoview.
+
+
+
 infoview.get_current_infoview()              *infoview.get_current_infoview()*
     Get the infoview corresponding to the current window.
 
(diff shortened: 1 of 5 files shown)


Julian:
I was doing the same thing locally :), you're faster at hitting the PR button. Will stop and let you deal with it given you've played with this longer, but yeah +1 on all the above.

Julian:
I'm also probably in favor of mapping CR to click, which I think you mentioned somewhere else that I can't find, though I see it's not currently mapped.

Julian:
blehhh sorry I am dumb and accidentally merged this whilst trying it

Julian:
(OK all clear, sorry for the noise.)

gebner:
Okay, there's one thing left on the todo list but I'll merge the PR now.

What's missing is the "show tooltip when moving the cursor over a term" feature (right now you still need to click).  From what I can tell this requires a bit of refactoring: 1) we should probably add events for entering/exiting a highlightable div. 2) these need to be hooked up so that they can cause a rerender (which I believe is handled with the `call_event` wrapper but I don't fully understand this part yet).

Julian:
Sounds good to me, thanks!

Issue #139, opened 2021-10-04
Additional widget tweak TODOs


Julian:
- [ ] Don't bind S / hop.nvim when hop.nvim isn't installed

rish987:
The `bad argument #1 to 'unpack' (table expected, got nil)` error messages have always been easy to properly fix when I've seen them. I'm just not sure how to reproduce this error, I've been trying at 73fedfd (where you said was the first place you got this) but no luck. Since you were able to bisect it, it must have pretty easy to reproduce, right?

Would you mind posting a screencast of the error (ideally a few runs so I can see the commonalities)?

rish987:
One more thing we'll want to mention is that the hop functionality is specifically dependent on phaazon/hop.nvim#153 (or they can use master on rish987/hop.nvim).

gebner:
> The infoview has an extra trailing newline at the bottom (which breaks G quite annoyingly)

Sorry, that's my fault.  I'd also love to get rid of the newline, but the tests failed otherwise. :shrug: 
https://github.com/Julian/lean.nvim/blob/373c2e269d2f7b2f69e124ce90833a1b7a233e79/lua/lean/infoview.lua#L262-L263

Julian:
> Sorry, that's my fault. I'd also love to get rid of the newline, but the tests failed otherwise. 🤷

:D all good -- though seems it passes now, just deleting those lines (in 75cee9ccbb77ecaeaf36c03ca1e4ad2f00ccab0a) seems to be all happy check marks, so away it goes...

PR #141, opened 2021-10-04
Better error handling in infoview

gebner:
The Lean 3 widget implementation had some issues (tried to serialize a lua function as json).  This caused neovim to raise an exception, which was not handled.  Due to the exception the lock in the tick remained locked and prevented further updates, causing the infoview to hang.

This PR fixes the function serialization issue and disables the locking in the ticks because I didn't find a reliable way to handle these fatal errors.

### DIFF
--- a/lua/lean/_util.lua
+++ b/lua/lean/_util.lua
@@ -107,6 +107,9 @@ end
 
 M.a_request = a.wrap(M.request, 4)
 
+-- FIXME: tick locking is disabled for now
+-- It is really easy to crash the infoview this way if an exception is not handled.
+
 local Tick = {}
 Tick.__index = Tick
 
@@ -118,10 +121,10 @@ function Tick:check()
   -- this tick has been cancelled
   if self.ticker.tick ~= self.tick then
     -- allow waiting ticks to proceed
-    if self.ticker._lock == self.tick then
-      self.ticker._lock = false
-      self.ticker.lock_var:notify_all()
-    end
+    -- if self.ticker._lock == self.tick then
+    --   self.ticker._lock = false
+    --   self.ticker.lock_var:notify_all()
+    -- end
     return false
   end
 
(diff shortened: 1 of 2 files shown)


Julian:
Two more checked off from #139 sounds like :) -- thanks again! LGTM.

lua/lean/lean3.lua
```diff
@@ -215,7 +215,8 @@ function lean3.update_infoview(pin, bufnr, params, use_widget, opts, options)
           local clickable_event = div_event == "click" or div_event == "change"
           if clickable_event then element_div.highlightable = true end
           events[div_event] = function(this_tick, value)
```
rish987, line 217: My bad! I should have made `value` the third parameter after I updated to be passing in the `render_fn`. Though this fixes it as well, right? So could we revert 2589253? (again, I'm a bit occupied this week so I may not be able to do much more that comment)

lua/lean/_util.lua
```diff
@@ -107,6 +107,9 @@ end
 
 M.a_request = a.wrap(M.request, 4)
 
+-- FIXME: tick locking is disabled for now
```
rish987, line 110: I think a better alternative to this is using a variant of `lsp.buf_request` that has a built-in timer after which it expires and calls the handler with an error (not sure if this already exists).

lua/lean/lean3.lua
rish987: (specifically, this was introduced in 27977d1)

lua/lean/_util.lua
gebner: That's what I tried first, but I couldn't get it working (without spending too much time on it).  For now, I think it's less annoying if the infoview is sometimes corrupted than if it stops updating completely.

rish987: Right, but doesn't 6d8314c fix it without this?

gebner: I've run into similar problems (infoview not updating) with Lean 4 as well, but they're harder to reproduce.

rish987: Interesting. And you're sure the the Lean 4 server is guaranteed to respond to every LSP request that we're making here? And when it did halt in Lean 4, did you get an error?

I'm just concerned that doing this -- in addition to the bugs it introduces -- serves to hide issues that we need to solve, here or possibly even in the server itself.

PR #146, opened 2021-10-05
Only set `data_div` at the very end.

### LINKED ISSUE #142: Moving the cursor over regions can garble the infoview with concatenated goal state
See this recording:

https://user-images.githubusercontent.com/329822/135878886-ab4ee5ca-18c7-4770-ad7b-c95c0f59d6e8.mov

where somehow holding down `l` (or `h`) and moving over the intermediate regions leaves the infoview with somehow multiple copies of the state once the cursor lands.

Moving the cursor again "fixes" / re-renders correctly.

rish987:
I'm trying to scratch my brain for the real reason I added the lock, I think it had something to do with the Lean 3 infoview being left in a stale state if a replay was interrupted, but I think the dummy div used when setting loading makes it unnecessary. So I suppose you can ignore all of my cries to revert 2589253364aa3142a0b52d7dcae9c71c1d32b539 and we can do without the lock for now by simply avoiding concurrently modifying `data_div` :P

Should close #142 by fixing the race condition that I think caused it, but I'm pretty pinched on time this week so more a bit more testing would be appreciated!

Also something unrelated I noticed is that the blue highlights don't work anymore? Not sure if this was intentional.

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -610,7 +610,7 @@ function Pin:__update(tick, delay, lean3_opts)
   delay = delay or 100
 
   self:set_loading(true)
-  self.data_div = html.Div:new({pin = self}, "", "pin-data", nil)
+  local new_data_div = html.Div:new({pin = self}, "", "pin-data", nil)
 
   if delay > 0 then
     util.wait_timer(delay)
@@ -620,7 +620,7 @@ function Pin:__update(tick, delay, lean3_opts)
 
   local buf = vim.fn.bufnr(vim.uri_to_fname(params.textDocument.uri))
   if buf == -1 then
-    self.data_div:insert_div({}, "No corresponding buffer found.", "no-buffer-msg")
+    error("No corresponding buffer found for update.")
     return false
   end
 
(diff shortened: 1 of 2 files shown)


gebner:
This approach feels more reliable than the lock.  It is less likely to get into an unrecoverable crashed state, compared to the lock which needs to be released manually (and where the release could be prevented by an exception).

> Lean 3 infoview being left in a stale state if a replay was interrupted

Why are we doing replays in Lean 3 in the first place?  The Lean 3 server keeps track of the widget state, so we shouldn't have to do anything there.

> Also something unrelated I noticed is that the blue highlights don't work anymore? Not sure if this was intentional.

That's probably my fault.  The highlights didn't work in the GUI at all (because only terminal colors were specified), so I replaced them by similar-looking links: https://github.com/Julian/lean.nvim/commit/d364e021f9b279e1f4d8226836d44ab542fb2a73  (I am not sure if hardcoding the colors is a good idea in general, because they need to work with different light and dark themes.)

I did test the new highlights but I probably missed something then.  Which highlight is broken for you?

rish987:
> Why are we doing replays in Lean 3 in the first place?

Undo has three steps:
1. Remove the last item from the undo stack
2. Clear everything by triggering any `clear` events until none are left
3. Replay the undo stack

This was more useful particularly in Lean 3 when tooltips were still inline.

gebner:
Oh wow.  It's amazing that you got this working at all.

Strategically, I think it's more useful to focus on Lean 4 functionality.  It's acceptable that Lean 3 undo is a bit brittle if it means that the Lean 4 infoview doesn't hang.

PR #148, opened 2021-10-05
Show stderr output in :messages

### LINKED ISSUE #37: Standard error output from the server should be shown
The Lean server prints many fatal and not-so-fatal errors on stderr (like stack overflows, panics, assertion failures etc.).  These should be shown somehow.

gebner:
Fixes #37.

### DIFF
--- a/lua/lean/stderr.lua
+++ b/lua/lean/stderr.lua
@@ -1,0 +1,20 @@
+local log = require'vim.lsp.log'
+local stderr = {}
+
+-- Show stderr output in :messages
+-- TODO: add upstream neovim API
+function stderr.enable()
+  local old_error = log.error
+  log.error = function(...)
+    local argc = select('#', ...)
+    if argc == 0 then return true end -- always enable error messages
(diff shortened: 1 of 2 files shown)


lua/lean/stderr.lua
```diff
@@ -0,0 +1,19 @@
...
+-- Show stderr output in :messages
+-- TODO: add upstream neovim API
+function stderr.enable()
+  local old_error = log.error
+  log.error = function(...)
+    local argc = select('#', ...)
+    if argc == 0 then return true end -- always enable error messages
+    if argc == 4 and select(1, ...) == 'rpc' and select(3, ...) == 'stderr'
+        and string.match(select(2, ...), 'lean') then
+      print(select(4, ...))
```
Julian, line 13: Perhaps `vim.api.nvim_err_writeln` instead of `print`?

gebner: Ok, changed!

Julian:
lgtm!

gebner:
Mmh, we could put the last ten lines in the infoview and then open a new buffer (with the full stderr output) on enter.

PR #150, opened 2021-10-05
Bind tab to enter tooltips.

Julian:
Perhaps in preparation for shift-tab moving backwards
through them?


gebner:
Sounds good to me, also the shift-tab binding.

PR #152, opened 2021-10-06
Fix for nil path error when first cursor position in the infoview is outside of cell boundaries

rish987:
@Julian does this fix the issue for you related to 4ee7e38? (It should be pretty obvious to you if it doesn't since I added error prints to the hack fix.)


rish987:
(oops, I mean to refer to ac22758, not 4ee7e38)

rish987:
I understand neither of you are very familiar with the Div code yet and so perhaps cannot judge this change -- but this is definitely a fix for an edge case I overlooked, so will merge if no one has any obvious issues.

And plus we have some new beta testers now ;)

Julian:
Seems like indeed after this I don't see the unpack errors, thanks!

PR #154, opened 2021-10-06
Bind LocalLeader+Tab to jump into the infoview.

Julian:
Useful particularly if there are lots of windows positioned in between
the current one and the infoview.


rish987:
Awesome! Thank you!

PR #155, opened 2021-10-06
Tests: initialize buffer/window tracking on `helpers.setup()`.

### LINKED ISSUE #153: require'tests.helpers' has side effects, which produce errors when used interactively
Simply importing the helper file appears to now have side effects, specifically it runs some code I think we use for some infoview window tests:

https://github.com/Julian/lean.nvim/blob/main/lua/tests/helpers.lua#L837-L839

Which means if I want to debug a (new) test by running:

`make nvim SETUP_TABLE='{ lsp3 = { enable = true }, mappings = true }'`

and then run `:edit foo.lean` and `:lua require'tests.helpers'` from within it, I get obtuse errors:

```
:lua require'tests.helpers'                                                                                                                                                                                                                                               0,0-1          All lean://info/1                   1,0-1          All
E5108: Error executing lua /Users/julian/Development/lean.nvim/lua/tests/helpers.lua:311: expected: {                                                                                                                                                                                                                 
  [1000] = true
}
 got: {
  [1000] = true,
  [1002] = true
}
Expected to be truthy, but value was:
(boolean) false
```

Removing those two lines makes the errors go away, but makes lots of tests fail.

rish987:
@Julian should close #153.

### DIFF
--- a/lua/tests/helpers.lua
+++ b/lua/tests/helpers.lua
@@ -34,6 +34,9 @@ local default_config = {
 }
 
 function helpers.setup(config)
+  -- initialize on very first nvim window/buffer (base case satisfied pretty trivially)
+  assert.win.created.tracked()
+  assert.buf.created.tracked()
   lean.setup(vim.tbl_deep_extend("keep", config, default_config))
 end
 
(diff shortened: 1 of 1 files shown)


Julian:
Hahaha that's way easier than even what I thought the easy thing was (splitting out all the stateful stuff into a separate module. Which seems also easy but waiting for the tests to run repeatedly is super slow ugh) -- well done, yeah let's go with this for now then.

Julian:
Sounds good to me, and thanks again for managing to fix this before I'd even diagnosed :D

PR #160, opened 2021-10-07
Bind <Esc> to close tooltips in infoview windows

Julian:
So -- `<tab>` jumps in, `<s-tab>` jumps outward, `<esc>` closes em all.


rish987:
Also, the way that this TUI works is if you're hovered over a div with a tooltip, that tooltip will show. If you want we can add a command that disables/enables this hover feature.

Julian:
I think I saw you mention the hover thing elsewhere -- it sounds like a good idea though I haven't used widgets for long enough yet to know whether I'd personally configure mine on or off -- but I can imagine that autohover is one angle (whether widgets automatically show any interactable widget underneath them as soon as you move the cursor on top of them), but regardless of that autohover setting I think sometimes you just want less noisiness for a second in the window -- my brain would mash the Esc key in that case, same as for exiting insert mode to get back to some calm, which is the impetus for suggesting this mapping -- but I think even if autohover was on, if someone hit Esc, the tooltip shouldn't re-autohover in my mind.

rish987:
> but I think even if autohover was on, if someone hit Esc, the tooltip shouldn't re-autohover in my mind
This should be how it works at the moment - I've made the clear commands specifically remove the tooltip from the div entirely, so when you press `C` you shouldn't see a tooltip at all. Are you seeing something else?

Or are you saying hypothetically in a div that you can't modify at all, in which case yeah we could add a way to disable hovers (in case you get overwhelmed by them, or just want to see underneath without having to move your cursor).

PR #161, opened 2021-10-08
Use hlgroups compatible with light theme pls.

rish987:
I'm a light theme user (I'm very sorry) and the recent changes to the hlgroups cause these to be indistinct from the background for me in the tooltips. Hopefully this is a good compromise.


rish987:
😆 So I think I'll leave it at this for now, but we definitely want to eventually add an easy way to configure these hlgroups, as there's no way we'll be able to please everyone...

PR #163, opened 2021-10-08
Fix for Lean 3 goto when going to definition in the same file.

rish987:


Julian:
I assume here goto definition is somehow a widget related goto definition?

Issue #162, opened 2021-10-08
Infoview does not update while processing


gebner:
When a file is still processing, the infoview only shows the unhelpful "Processing file..." message.  It should show the available diagnostics and tactic state for already-processed regions.

rish987:
From #165:
> AFAICT it's an upstream Lean 4 server issue. It seems to not be responding to our Lean.Widget.getInteractiveDiagnostics RPC calls until the file is fully processed, so many of our async calls are left hanging indefinitely.

One workaround we could use is to put a timer on `util.a_request` after which it calls the handler with error, and let it fall back to the plain diagnostics. But a server fix, at least to respond error rather than not respond at all, is certainly needed.

gebner:
Thanks for debugging that!  Would it be enough if the getInteractiveDiagnostics command would return once Lean is done with the corresponding line (but not necessarily the whole file)?  Or would the infoview still get stuck?

rish987:
We don't send the request at all if we detect the line is processing. So as long as it's responsive after `fileProgress` indicates that line is done processing, that will work. Though ideally it seems it should always be responsive regardless.

PR #165, opened 2021-10-08
Show processing message for lean 3 as well.

rish987:
@gebner I didn't notice #162 with this -- when I opened up a big Lean 3 file, it didn't get stuck on "Processing file..." when the orange bars were gone in a particular range, while other parts were still processing. Although the infoview was empty there, I'm guessing because the response was empty.

What's a good file to test this on for Lean 4? Or where in particular did you see the problem?


rish987:
Correction: actually, the infoview does in fact get me messages while parts of the file are still processing.

PR #167, opened 2021-10-11
Define :commands.

gebner:
This makes it much nicer to set up custom keybindings.  I've also taken the liberty to fix some corner cases with the infoview commands (e.g. toggle didn't reopen it if you closed it with `:q`).

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -109,6 +109,8 @@ end
 
 --- Open this infoview if it isn't already open
 function Infoview:open()
+  if self.is_open then return end
+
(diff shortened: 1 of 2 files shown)


rish987:
> I'll remove the __maybe_autoopen calls. If I understand the invariant correctly, every lean(3) buffer should have a get_current_infoview, right?

Good question -- `autocmd BufEnter ...` will trigger it, so before it was unnecessary since all the mappings to these calls were buffer-local. But now it looks like you want them to be able to be called from anywhere? So they may in fact be necessary at this point...

PR #169, opened 2021-10-11
Add lineRange support to getInteractiveDiagnostics

### LINKED ISSUE #162: Infoview does not update while processing

gebner:
Pending Lean PR https://github.com/leanprover/lean4/pull/722

Fixes #162.

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -707,7 +707,7 @@ function Pin:__update(tick, delay, lean3_opts)
 
     local diagnostics_div
     if self.use_widget then
-      local diags, err = self.sess:getInteractiveDiagnostics()
+      local diags, err = self.sess:getInteractiveDiagnostics({ start = line, ['end'] = line + 1 })
       if not tick:check() then return true end
       if not err then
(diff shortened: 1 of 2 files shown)


gebner:
It looks like the documentation generator gave up on this PR.  Not sure how to fix this.

(Personally, I don't get anything out of the auto-generated documentation.  Even more, I find the bot commits on the PRs very annoying.  As far as I am concerned, we could just drop the `lean.txt` or replace it by the brief explanation at the top of `init.lua`.)

Julian:
I'm fine with dropping it, though I do find it annoying that `:h lean.txt` doesn't have more useful info in it even for myself -- it's what a native vim user would expect for docs. The autogeneration was just a hope we wouldn't have to spend more time manually doing it, but it doesn't seem we're getting a lot of value there to me either so it's fine if we kill it and give up for now I guess.

Julian:
We could of course alternatively just have the bot run on master and not PRs.

PR #174, opened 2021-10-13
Fix lean.current_search_paths() now that Lean4 returns JSON from leanpkg print-paths

Julian:


Julian:
OK it does for me now locally but not in CI... will have to see why.

PR #180, opened 2021-10-15
Div code refactor: always attempt to restore div position, even when not in the window.

rish987:
The main use case of this is for pins. In particular, if you have gone into a stack of tooltips in a (non-current) pin, and then return to the source and update the current pin, it will attempt to preserve your position so that 
1. the same stack of tooltips will still be visible, and
2. when you return (using `infoview.go_to()`) you'll be in the same position in that tooltip stack.

Previously, we only did this if you were actually in the infoview window itself (which is how you were able to stay in the same tooltip after a click triggered a re-render).

Since this changes quite a bit of the core div code, wanted to at least test this out a bit here to make sure there aren't any obvious problems/regressions before merging.

### DIFF
--- a/lua/lean/html.lua
+++ b/lua/lean/html.lua
@@ -568,14 +568,17 @@ end
 
 function Div:buf_update_position(buf)
   local bufdata = self.bufs[buf]
-  local raw_pos = pos_to_raw_pos(vim.api.nvim_win_get_cursor(bufdata.last_win),
-    vim.api.nvim_buf_get_lines(buf, 0, -1, true))
+  local cursor_pos = vim.api.nvim_win_get_cursor(bufdata.last_win)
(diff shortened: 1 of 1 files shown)


rish987:
One thing to take note of: because of a neovim issue which I think is related to neovim/neovim#13403, we have to do a `:redraw` after creating the tooltip windows. So while moving around in the source, you'll probably see an amount of cursor flickering proportional to the number of tooltips you have open in the infoview.

Julian:
This causes uncontrolable flickering for me.

See the video:

https://user-images.githubusercontent.com/329822/137604437-de20dbd4-5f32-4a53-a7d4-d42e2587e73e.mov

That seems logically more related to #182 which I see is combined in this PR, probably a good reason to keep them separate.

rish987:
Sorry, should have put the merge with #182 on a different branch. Does it happen at 93af6e8?

Issue #181, opened 2021-10-15
Performance issues with long traces in infoview


gebner:
When the infoview shows a long trace, scrolling through the infoview (with j and k) becomes extremely slow and takes 100% CPU.  This is a good Lean 4 file to try out:
```lean
def Good (n : Nat) := True

@[simp] theorem Good_zero : Good 0 := ⟨⟩
@[simp] theorem Good_succ_iff : Good (n+1) ↔ Good n := Iff.rfl

set_option trace.Meta.Tactic.simp true in
example : Good 500 := by simp
```

PR #182, opened 2021-10-16
Throttle div-related UI updates.

rish987:
@gebner does this help with #181? (there are of course things I could optimize in the TUI code but I'd rather focus on correctness at the moment)


rish987:
@Julian are you getting the flickering tooltip here? I've not been able to reproduce this myself...

PR #183, opened 2021-10-18
Optimization: avoid re-renders where possible.

rish987:
@gebner Took your advice from #182. This should help "halfway" with #181 in the sense that it  won't rerender if it doesn't have to. In particular for the test file there if you go to the beginning of the line (`0`) and start scrolling with `j` or `k` it should be fast because there are no interactive elements there. However, if you got to the end of the line (`$`) and scroll it will be slow because it will be repeatedly entering/exiting elements with interactive decorations (and so will have to re-render).

### DIFF
--- a/lua/lean/html.lua
+++ b/lua/lean/html.lua
@@ -434,14 +434,17 @@ function Div:buf_render(buf, internal)
   local text, hls = root:render()
   local lines = vim.split(text, "\n")
 
-  vim.api.nvim_buf_set_option(buf, 'modifiable', true)
-  vim.api.nvim_buf_set_lines(buf, 0, -1, true, lines)
-  -- HACK: This shouldn't really do anything, but I think there's a neovim
-  --       display bug. See #27 and neovim/neovim#14663. Specifically,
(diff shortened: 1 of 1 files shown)


rish987:
Also prevented setting all the buffer lines unnecessarily on internal renders, but this didn't seem to be enough to fix the performance issue above. So that suggests that Lua is indeed the bottleneck here.

gebner:
Great!  This is already a significant improvement.  Interestingly enough, scrolling at the end of the infoview (line 1500 in the example) is much slower than at the beginning.

> So that suggests that Lua is indeed the bottleneck here.

I'm confident that we can make this fast, and I don't think Lua is going to be the issue there.  When you move the cursor, we only need to do one thing: check if there's a tooltip and maybe toggle the highlighting.

None of that is necessarily slow.  Figuring out whether there's a tooltip could use some optimizations (maybe store the range of each div so that we can skip whole subtrees?).  Toggling the highlighting might require some restructuring, i.e. rendering the highlighting separately and not part of the div tree.

PR #184, opened 2021-10-20
Show current pin extmark when inside infoview.

rish987:
Just for a bit more context while you're poking around in there :)


Julian:
This is a bit noisy to me personally -- both of you like seeing this even when the infoview isn't in debug mode?

PR #191, opened 2021-10-31
Fix Lean 4 click on non-clickable divs inside tooltips closing the tooltip.

rish987:
@gebner Here's what I had in mind from the discussion at #190.


gebner:
This is just a complicated way to write `div.events.click = function() end`, right?

PR #193, opened 2021-11-08
Make things in _target nomodifiable by default.

Julian:
Add a config option to turn that off, but the default should prevent accidental modification of dependencies.
    
See https://leanprover.zulipchat.com/#narrow/stream/113488-general

Doesn't add Lean4 support since I haven't used Lake personally yet and don't know where it stores stuff, but happy to add that (now or later depending on how easy it is).


lua/lean/ft.lua
```diff
@@ -45,6 +51,15 @@ function ft.detect(filename)
   vim.opt.filetype = filetype
 end
 
+--- Maybe make the current buffer `nomodifiable` unless configured not to.
+---
+--- Prevents unintentional modification of files in `_target` (i.e. Lean
+--- dependencies).
+function ft.__maybe_make_nomodifiable(fname)
+  if not options.make_deps_nomodifiable then return end
+  if fname:find('/_target/.*/.*.lean') then vim.bo.modifiable = false end
```
rish987, line 60: Could we do the same thing for standard library files?

Julian: Yes def good idea, moving this to draft since I had the same thought yesterday (I'm gonna change this to instead be a configurable set of patterns).

PR #195, opened 2021-11-09
Use separate buffer for stderr messages.

gebner:
See also #148.  Printing stderr to `:messages` is extremely annoying when there's a constant stream of messages (such as when the server panics on certain syntax, like https://github.com/leanprover/lean4/issues/776) because `:messages` steals the focus and you can't type anymore.  This PR insteads opens a new `lean://stderr` buffer and prints the stderr output there.

I've thought about putting the stderr output in the infoview as discussed in #148 but I don't think that's a good idea since we want to reduce the amount of infoview updates (as in #190).

The buffer is disabled in the tests because it also shows that elan is downloading a new lean version (which is on stderr and happens in the tests).

### DIFF
--- a/lua/lean/stderr.lua
+++ b/lua/lean/stderr.lua
@@ -1,17 +1,61 @@
 local log = require'vim.lsp.log'
+local infoview = require'lean.infoview'
 local stderr = {}
 
--- Show stderr output in :messages
+-- Opens a window for the stderr buffer.
+local function open_window(stderr_bufnr)
+  local old_win = vim.api.nvim_get_current_win()
+
+  -- split the infoview window if open
+  local iv = infoview.get_current_infoview()
+  if iv and iv.window and vim.api.nvim_win_is_valid(iv.window) then
+    vim.api.nvim_set_current_win(iv.window)
+    vim.cmd(('rightbelow sbuffer %d'):format(stderr_bufnr))
+  else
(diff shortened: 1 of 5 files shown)


lua/lean/stderr.lua
```diff
@@ -1,17 +1,43 @@
...
-      vim.schedule(function() vim.api.nvim_err_writeln(chunk) end)
+      vim.schedule(function()
+        if not stderr_bufnr or not vim.api.nvim_buf_is_valid(stderr_bufnr) then
+          stderr_bufnr = vim.api.nvim_create_buf(false, true)
+          vim.api.nvim_buf_set_name(stderr_bufnr, "lean://stderr")
+          stderr_winnr = nil
+        end
+        if not stderr_winnr or not vim.api.nvim_win_is_valid(stderr_winnr) then
+          local old_win = vim.api.nvim_get_current_win()
+          vim.cmd(('botright sbuffer %d'):format(stderr_bufnr))
```
Julian, line 23: I'd slightly prefer always splitting this below the infoview window (i.e. not having it be full-width) which at least for the typical window layout will cause less jittering. 

Don't feel strongly though if you like it this way.

gebner: This might be a good idea.  I'll try it out.

gebner:
Rebased to main and it splits the infoview now instead of the whole window.

Julian:
Nice!

Issue #166, opened 2021-10-11
Automatically show tooltip when moving over term


gebner:
Didn't manage to add this in #138:

> What's missing is the "show tooltip when moving the cursor over a term" feature (right now you still need to click). From what I can tell this requires a bit of refactoring: 1) we should probably add events for entering/exiting a highlightable div. 2) these need to be hooked up so that they can cause a rerender (which I believe is handled with the call_event wrapper but I don't fully understand this part yet).

PR #196, opened 2021-11-10
Implement cursor events and use to get Lean 3 click higlighting.

rish987:
This is very useful for precedence hints. We technically already had this with the `I`/`i` keybindings but this makes it automatic. @gebner this should also make it easy to implement #166 (though I'd like this to be configurable).

### DIFF
--- a/lua/lean/html.lua
+++ b/lua/lean/html.lua
@@ -250,7 +250,7 @@ end
 ---@class DivEventContext
 ---@field rerender fun()
 ---@field rehover fun()
----@field root BufDiv
+---@field self BufDiv
 
 ---Trigger the given event at the given path
 ---@param path PathNode[] @the path to trigger the event at
--- a/lua/lean/html.lua
+++ b/lua/lean/html.lua
@@ -491,7 +491,7 @@ function BufDiv:buf_update_position()
   self.path = self.div:path_from_pos(raw_pos)
 
   if not path_equal(path_before, self.path) then
-    self:buf_event("cursor_leave", path_before, function() self:buf_event("cursor_enter") end)
+    self:buf_event("cursor_leave", path_before)
     self:buf_event("cursor_enter", self.path)
     return true
   end
@@ -612,7 +612,7 @@ function BufDiv:buf_make_event_context()
   return {
     rerender = function() self:buf_render() end,
     rehover = function() self:buf_hover() end,
-    root = self,
+    self = self,
   }
 end
 
--- a/lua/lean/html.lua
+++ b/lua/lean/html.lua
@@ -490,7 +490,13 @@ function BufDiv:buf_update_position()
 
   self.path = self.div:path_from_pos(raw_pos)
 
-  return not path_equal(path_before, self.path)
+  if not path_equal(path_before, self.path) then
+    self:buf_event("cursor_left", path_before, function() self:buf_event("cursor_entered") end)
+    self:buf_event("cursor_entered", self.path)
+    return true
+  end
+
+  return false
 end
 
 function BufDiv:buf_hover(force_update_highlight)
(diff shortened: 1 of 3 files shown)


Julian:
Can you quickly describe what this does? I can try to test it out and see. From a guess, is this automatic clicking on things as you move into them?

rish987:
> is this automatic clicking on things as you move into them?

No, that would be #166. This gets highlighting from the server that shows you what you would be clicking on. Try hovering over a binary operator, for example.

lua/lean/infoview.lua
```diff
@@ -503,7 +501,7 @@ function Pin:set_loading(loading)
   return false
 end
 
-Pin.update = a.void(function(self, force, delay, _, lean3_opts)
+function Pin:a_update(force, delay, _, lean3_opts)
```
Julian, line 504: Can you name this something a bit more transparent? What is `a`? Is that short for async?

lua/lean/lean3.lua
```diff
@@ -49,8 +49,8 @@ local class_to_hlgroup = {
 
 -- mapping from lean3 events to standard div events
 local to_event = {
-  ["onMouseEnter"] = "mouse_enter";
-  ["onMouseLeave"] = "mouse_leave";
+  ["onMouseEnter"] = "cursor_entered";
```
Julian, line 52: I think personally I very slightly prefer the older names, since below when you use `cursor_left`, in isolation it's easy to confuse that for something that's the opposite of `cursor_right`.

Julian:
Got it, thanks.

lua/lean/html.lua
```diff
@@ -612,7 +612,7 @@ function BufDiv:buf_make_event_context()
   return {
     rerender = function() self:buf_render() end,
     rehover = function() self:buf_hover() end,
-    root = self,
+    self = self,
```
gebner, line 615: BTW, this is no longer type-correct.  Also, the `clear_all` handler uses the `root` field.

rish987: Whoops, my bad, fixed for Lean 4. What do you mean by it not being type-correct?

gebner: You've already fixed it.  The type declaration for `DivEventContext` had a `root` field which wasn't provided here.

(Type checking support in the sumneko language server is still a work in progress and not enabled by default yet afaict.)

Julian: It's also somewhat annoying to wire up in CI which I gave a quick look at. Didn't see an easy emmy type checker to use though.

gebner:
> this should also make it easy to implement #166 (though I'd like this to be configurable).

I'm not sure what the best behavior there is either.  I only know that pressing enter and then tab is cumbersome, and then it's weird that enter isn't necessary the second time.  Maybe tab should open the tooltip as well, and the tooltip doesn't reappear after it is closed.  I've been meaning to write up some UX proposals for #166, I'm not dead set on having tooltips appear automatically.

lua/lean/html.lua
gebner: The only other emmy typechecker I know of is https://github.com/Benjamin-Dobell/IntelliJ-Luanalysis/, which would be even harder to run in CI.  We can probably contribute a PR for a command-line linter mode if we want to integrate the sumneko checker in CI.

Julian:
> I've been meaning to write up some UX proposals for #166

Please do because I'm not 100% sure I feel comfortable with the current UX either (by which I mean I'm not sure what I want yet even -- so idea generation will be useful)

rish987:
> I only know that pressing enter and then tab is cumbersome, and then it's weird that enter isn't necessary the second time

This is a non-issue for me since I've been using the hop.nvim integration, so I don't even use tab or J, I just jump almost instantly to wherever I want. But I'd like this to be configurable because I wouldn't want the tooltip to show up and cover the rest of the text unless I specifically click on something.

PR #197, opened 2021-11-10
Add notes on hop.nvim support and make it configurable.

rish987:
I'll be incrementally PR'ing parts of phaazon/hop.nvim#153, but not sure how long this is going to take. Since my fork has been working perfectly for me for quite some time now, I think it's at least worth a mention at this point.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,6 +36,7 @@ Install via your favorite plugin manager. E.g., with
     Plug 'hrsh7th/nvim-compe'  " For LSP completion
     Plug 'hrsh7th/vim-vsnip'   " For snippets
     Plug 'andrewradev/switch.vim'  " For Lean switch support
+    Plug 'rish987/hop.nvim'  " TEMPORARY: for experimental interactive infoview hop support (to be eventually merged into phaazon/hop.nvim)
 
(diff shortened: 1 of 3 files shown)


rish987:
Also, ggandor/lightspeed.nvim seems to be getting a little more action, I may eventually just go with that instead.

Julian:
I'd slightly prefer waiting till it's merged upstream, given support is really here so it's just about churning the README to first link to the fork and later to upstream, and in between stuff can change which might annoy someone else who assumes support is stable. Especially so if you intend to rewrite to use the other implementation. Feel free to insist if you really want to merge this, I don't think it matters terribly much, but yeah slight preference to close and wait given support is anyways there if someone's adventurous and wants to try it.

Also implementation-wise, the way some of the other optional functionality works is more like that the default is true if you have the thing installed, and false if it isn't, e.g. switch or vsnip or treesitter basically turns on if you have those plugins but does nothing if you don't, which I think is the right pattern -- if someone has hop (or easymotion or whatever other plugin) installed it should default on but be disablable, but if someone does not have hop installed, it should default to off.

Julian:
Closing as per the above but yeah feel free to follow-up if you want.

PR #198, opened 2021-11-11
Add support for Lean scratch buffers.

rish987:
This is a companion to #193. Sometimes I want to mess around with a Lean 3 file without causing all of the dependent files to have to reload. So what this does is just copy the text into a modifiable file with the same name and the postfix "_scratch.lean". Let me know if there's some other way I'm unaware of.

### DIFF
--- a/lua/lean/init.lua
+++ b/lua/lean/init.lua
@@ -92,6 +92,24 @@ function lean.is_lean_buffer()
   return filetype == "lean" or filetype == "lean3"
 end
 
+function lean.open_scratch_buffer()
+  if not lean.is_lean_buffer() then return end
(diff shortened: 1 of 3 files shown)


Julian:
`:sav %:r_scratch.lean` is the vim way to do this, I don't think we need a plugin/Lean-specific one.

rish987:
Thanks, that's almost what I want. One problem is that it immediately saves the file. Do you know if there's a similar command that opens a `setlocal buftype=nowrite` buffer instead?

Julian:
There's `:file` if you don't want to write it. I don't know if there's something that automatically sets `nowrite` but `file %:r_scratch.lean | setlocal buftype=nowrite` should work obviously if that's the behavior you're looking for.

Issue #194, opened 2021-11-08
Implement support for lake print-paths


Julian:
Currently we call `leanpkg print-paths` on Lean 4 in one place -- this should be replaced with the new Lake (JSON?) `print-paths` command.

PR #200, opened 2021-11-11
Add support for Lake to current_search_paths.

Julian:
Support for leanpkg.toml files for Lean 4 is for now left in place (but
only is used when there's no lakefile.lean).

Closes: #194

Refs: neovim/nvim-lspconfig#1390

### DIFF
--- a/lua/lean/init.lua
+++ b/lua/lean/init.lua
@@ -103,10 +103,16 @@ function lean.current_search_paths()
     paths = require'lean.lean3'.__current_search_paths()
   else
     local root = util.list_workspace_folders()[1]
-    -- print-paths emits a colon-separated list of .lean paths on the second line
+    if not root then root = vim.fn.getcwd() end
+
+    local executable = (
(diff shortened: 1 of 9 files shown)


lua/lean/init.lua
```diff
@@ -103,10 +103,16 @@ function lean.current_search_paths()
     paths = require'lean.lean3'.__current_search_paths()
   else
     local root = util.list_workspace_folders()[1]
-    -- print-paths emits a colon-separated list of .lean paths on the second line
+    if not root then root = vim.fn.getcwd() end
```
Julian, line 106: This fixes a small side issue which is this now works even if you don't have a lean buffer open, it defaults then to the obvious cwd.

rish987:
Thanks. Just FYI to both of you I may not be able to review much Lean 4 stuff for a bit since I need to focus on Lean 3 for my thesis project. So feel free to take over in that regard. (also PR # 200, nice)

Issue #173, opened 2021-10-13
Infoview cannot be reopened


gebner:
I can reproduce this with both `se hid` and `se nohid`:

 - `:e some_file.lean`
 - `:q`
 - (Now you're in the infoview.)
 - `:e another_file.lean`
 - `:LeanInfoviewToggle`

Then I get the error `infoview.lua:150: Vim:E444: Cannot close last window`.  Apparently the infoview believes it is still open.

Julian:
I don't think this is actually fixed, I still see the same behavior (and the pending test I just tried to re-enable still fails). @gebner lemme know if I'm missing something.

gebner:
Indeed, the bug still seems to be there.  I though I'd fixed it by checking `vim.api.nvim_win_is_valid` first, but that clearly wasn't the issue.

Julian:
(OK, now it's fixed, with a one word change.)

PR #201, opened 2021-11-15
Implement state diffs.

rish987:
The way this works is you can place down a "diff pin" which is diff'ed against the current info in a separate window using vim's built-in diff functionality. This makes it much easier to figure out exactly what changed from one state to the next. Demo:

[asciicast](https://asciinema.org/a/pefTLbP2PlVLZplCVxv25hxRg)

Widgets still somewhat work in this diff-mode, but there are a few problems with it, for example positioning of floating windows can be broken if the diff is showing deleted lines above. So in this case, for now you should just clear the diff pin if you're serious about interacting with the infoview.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -172,6 +172,8 @@ In Lean Files
 +------------------------+----------------------------------------------------+
 | ``<LocalLeader>dc``    | clear current infoview diff pin                    |
 +------------------------+----------------------------------------------------+
+| ``<LocalLeader>dd``    | toggle auto diff pin mode                          |
++------------------------+----------------------------------------------------+
 | ``<LocalLeader>s``     | insert a ``sorry`` for each open goal              |
 +------------------------+----------------------------------------------------+
 | ``<LocalLeader>t``     | replace a "try this:" suggestion under the cursor  |
--- a/README.rst
+++ b/README.rst
@@ -168,6 +168,10 @@ In Lean Files
 +------------------------+----------------------------------------------------+
 | ``<LocalLeader>c``     | clear all current infoview pins                    |
 +------------------------+----------------------------------------------------+
+| ``<LocalLeader>dx``    | place an infoview diff pin                         |
++------------------------+----------------------------------------------------+
+| ``<LocalLeader>dc``    | clear current infoview diff pin                    |
++------------------------+----------------------------------------------------+
 | ``<LocalLeader>s``     | insert a ``sorry`` for each open goal              |
 +------------------------+----------------------------------------------------+
 | ``<LocalLeader>t``     | replace a "try this:" suggestion under the cursor  |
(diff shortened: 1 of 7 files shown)


Julian:
Thanks, this definitely is functionality I was also interested in. I'm still however left feeling the implementation looks more complex than I'd expect -- specifically I still expect if we did not have the concept of a pin, and instead used vim's normal concept of windows, that this functionality would require extremely minimal amounts of code, and be close to just `:diffthis` run on two buffers (one a live infoview and one a dead/paused one), and not require any internal code at all (i.e. we'd just tell people to run that as one usually would in vim, and no special support should be needed).

I'll review the implementation and see if I notice any concrete places to incrementally make progress there, but I don't have high hopes yet -- perhaps in the meanwhile, can you perhaps make lots of the implementation here private? I'd rather not proliferate more Lua API though I doubt it'll be relied on quite yet.

Julian:
(#5 is also relevant, though what I wanted to implement there is even a layer on top of this where we automatically load a bunch of diff pins in this terminology all into the quickfix list, by successively diffing a bunch of statements one after the other. Obviously what's here gets us closer, which is nice.)

rish987:
> I still expect if we did not have the concept of a pin, and instead used vim's normal concept of windows, that this functionality would require extremely minimal amounts of code, and be close to just `:diffthis` run on two buffers (one a live infoview and one a dead/paused one)

I don't think it's that simple -- if I have a mapping that, say, inserts `set_option pp.all` near the top of the file, I'd expect the pins in both diff windows to be updated accordingly. I'd also want an extmark indicating where I'm diffing with, which pins already conveniently provide.

> can you perhaps make lots of the implementation here private?

Is d0c7107 what you had in mind?

> #5 is also relevant, though what I wanted to implement there is even a layer on top of this where we automatically load a bunch of diff pins in this terminology all into the quickfix list, by successively diffing a bunch of statements one after the other.

I've just added "auto diff pin" mode, which makes the diff pin automatically track the last cursor position. So if you're precise enough about your cursor movements, you'll immediately be able to see the diff with the last position. This already solves the "successive" use case for me, though #5 does seem interesting.

Julian:
> if I have a mapping that, say, inserts set_option pp.all near the top of the file, I'd expect the pins in both diff windows to be updated accordingly. I'd also want an extmark indicating where I'm diffing with, which pins already conveniently provide.

Yeah, I guess this is where we differ, I'd indeed prefer neither of these work, and instead have a simpler implementation. But let's see. I suspect we'll revisit as soon as one of us looks at #173 or similar issues.

> Is c7f84f4 what you had in mind?

Awesome, yep!

This is probably lgtm now, will take one more quick look and then do so, thanks as usual!

gebner:
Just wondering: from what I can tell this uses the built-in text-based diff support.  How hard would it be to do semantic diffs using the widget data structures?

rish987:
> How hard would it be to do semantic diffs using the widget data structures?

That's an interesting question, and probably something we eventually want to do. Not sure how much we can retrofit for Lean 3, but if we can get enough information from the server to exactly correlate the divs, then we should technically be able to insert the diff highlights/text ourselves and then just set `scrollbind` in the two windows (instead of `:diffthis`).

PR #204, opened 2021-11-17
Obtain lsp.client object for RPC.

### LINKED ISSUE #189: RPC broken after LspRestart
It seems that the RPC module continues to send keep-alive messages after executing `:LspRestart`, resulting in lots of errors like `Cannot send message to unknown document ...`

gebner:
This is one commit of #118 which didn't get merged in #123.

Probably fixes #189.

### DIFF
--- a/lua/lean/rpc.lua
+++ b/lua/lean/rpc.lua
@@ -6,9 +6,8 @@ local control = require'plenary.async.control'
 ---@class RpcRef
 
 ---@class Session
----@field bufnr number
+---@field client any vim.lsp.client object
 ---@field uri string
----@field closed boolean
 ---@field connected boolean
 ---@field connect_err any | nil
 ---@field on_connected function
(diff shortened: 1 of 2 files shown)


lua/lean/rpc.lua
```diff
@@ -18,50 +17,63 @@ local control = require'plenary.async.control'
...
       })
     end
   end))
   -- Terminate RPC session when document is closed.
   vim.api.nvim_buf_attach(bufnr, false, {
-    on_reload = function() self.closed = true end,
-    on_detach = function() self.closed = true end,
+    on_reload = function() self:close_without_releasing() end,
+    on_detach = function() self:close_without_releasing() end,
   })
```
rish987, line 48: ISTM that the fix to #189 consists of these lines? I wanted to ask why using the client directly is necessary when `vim.lsp.buf_notify` is designed to avoid the need for this, and in particular it seems very unlikely that a different client will have a `$/lean/rpc/keepAlive` handler. Seems a tad more readable to just pass in the buffer, but this is minor so feel free to go ahead if you're more comfortable with this.

gebner: I think the main fix is checking `is_stopped()` to figure out if the old server is stopped, the handlers here are only called when the buffer is reloaded.  This doesn't happen with `:LspRestart` iirc.  I've also fixed all the other places where the session is closed, and stopped the keepalive timer there as well.

> I wanted to ask why using the client directly is necessary [...]

It's more for clarity.  We don't want to send the requests to any server supporting the request, but they need to be sent to exactly to the server we've connected to.  Other servers (or even a restarted server) won't know about the session id, the RPC references, etc.  Using the client instead of the buffer number expresses this intention.

rish987: OK, sounds good to me.

PR #208, opened 2021-11-21
Fix for UTF16 vs. UTF32 positions.

### LINKED ISSUE #206: Hover positions are incorrect for lines with multibyte characters
I'm about to head out for the evening, so this may be an easy fix, but the infoview seems to break totally if in Lean 3 I have a line like:

```lean
variables {𝔽 : Type*} {
```

Save that to a file, open it, and just append to the end of the line, and the traceback says:

```
Error detected while processing CursorMovedI Autocommands for "<buffer=1>":                                                                                                                     
E5108: Error executing lua /Users/julian/Development/lean.nvim/lua/lean/infoview.lua:545: index out of range                                                                                    
stack traceback:                                                                                                                                                                                
        [C]: in function 'str_byteindex'                                                                                                                                                        
        /Users/julian/Development/lean.nvim/lua/lean/infoview.lua:545: in function 'update_extmark'                                                                                             
        /Users/julian/Development/lean.nvim/lua/lean/infoview.lua:531: in function 'set_position_params'                                                                                        
        /Users/julian/Development/lean.nvim/lua/lean/infoview.lua:624: in function 'move'                                                                                                       
        /Users/julian/Development/lean.nvim/lua/lean/infoview.lua:443: in function 'move_pin'                                                                                                   
        /Users/julian/Development/lean.nvim/lua/lean/infoview.lua:846: in function '__update'                                                                                                   
        [string ":lua"]:1: in main chunk                                                                                                                                                        
```

(This obviously occurs when that line is in the middle of a longer file as well, and persists, so it basically just breaks the infoview).

It's still odd to me to see anything talking about pin updating when no pins are being used -- it seems something is miscalculating a character (which may be an easy fix), but in general I'd love it if no pin updates occurred unless I try to use pins.

More generally we probably should wrap infoview update calls in `pcall` and catch any failures and render them into the infoview so at least they don't repeatedly disrupt typing.

rish987:
Closes #206 following neovim/neovim#16382, though should be backwards-compatible. Will add some proper tests soon.

### DIFF
--- a/lua/lean/_util.lua
+++ b/lua/lean/_util.lua
@@ -236,4 +236,22 @@ end
 
 M.wait_timer = a.wrap(function(timeout, handler) vim.defer_fn(handler, timeout) end, 2)
 
+--- Utility function for getting the encoding of the first LSP client on the given buffer.
+---@param bufnr (number) buffer handle or 0 for current, defaults to current
+---@returns (string) encoding first client if there is one, nil otherwise
+function M._get_offset_encoding(bufnr)
+  local offset_encoding
+
+  for _, client in pairs(vim.lsp.buf_get_clients(bufnr)) do
+    local this_offset_encoding = client.offset_encoding or "utf-16"
+    if not offset_encoding then
+      offset_encoding = this_offset_encoding
+    elseif offset_encoding ~= this_offset_encoding then
+      vim.notify("warning: multiple different client offset_encodings detected for buffer, this is not supported yet", vim.log.levels.WARN)
+    end
+  end
+
+  return offset_encoding
+end
+
 return M
--- a/lua/lean/_util.lua
+++ b/lua/lean/_util.lua
@@ -202,20 +202,36 @@ function M.list_workspace_folders()
   return workspace_folders
 end
 
+---@class UIParams
+---@field filename string
+---@field row number
+---@field col number
+
+--- Checks that the given position parameters are valid given the buffer they correspond to.
+---@param params UIParams @parameters to verify
+---@return boolean
 function M.position_params_valid(params)
-  local bufnr = vim.fn.bufnr(vim.uri_to_fname(params.textDocument.uri))
+  local bufnr = vim.fn.bufnr(params.filename)
   if bufnr == -1 then return false end
 
   local lines = vim.api.nvim_buf_get_lines(bufnr, 0, -1, true)
 
-  local line = params.position.line + 1
-  local col = params.position.character + 1
+  local line = params.row + 1
+  local col = params.col + 1
 
   if line > #lines then return false end
 
-  if col > vim.fn.strwidth(lines[line]) then return false end
+  if col > #lines[line] then return false end
 
   return true
+end
+
+function M.make_position_params()
+  local buf = vim.api.nvim_win_get_buf(0)
+  local row, col = unpack(vim.api.nvim_win_get_cursor(0))
+  row = row - 1
+
+  return { filename = vim.api.nvim_buf_get_name(buf), row = row, col = col }
 end
 
 M.wait_timer = a.wrap(function(timeout, handler) vim.defer_fn(handler, timeout) end, 2)
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -676,7 +676,7 @@ function Pin:update_position()
 
   local extmark_pos = vim.api.nvim_buf_get_extmark_by_id(buf, extmark_ns, extmark, {})
 
-  local encoding = vim.lsp.util._get_offset_encoding and vim.lsp.util._get_offset_encoding(buf) or "utf-32"
+  local encoding = util._get_offset_encoding(buf) or "utf-32"
   local use_utf16 = encoding == "utf-16"
   local new_pos = {}
 
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -59,6 +59,7 @@ local options = {
 ---@field data_div Div
 ---@field div Div
 ---@field ticker table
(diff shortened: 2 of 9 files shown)


lua/lean/_util.lua
```diff
@@ -218,6 +218,12 @@ function M.position_params_valid(params)
   return true
 end
 
+function M.make_position_params()
+  -- UTF-32 is used for general UI features and Lean 3 LSP; UTF16 is used for Lean 4 LSP
```
Julian, line 222: What kind of general UI features use this? Are they not limited to Lean 3?

I'd love to limit all of this encoding manipulation to just Lean 3 -- and to just having a separate `make_position_params` for Lean 3, rather than carrying around both encodings and having lots of places choose one or the other, that seems easy to make a mistake on.

Julian:
@rish987 just want to make sure you're not spinning wheels. Have you seen https://github.com/Julian/lean.nvim/pull/208#discussion_r753808471 ? This PR still seems like it's proceeding with something that can be limited to Lean 3.

rish987:
The way that I see it now is that we have parameters that are used for LSP update requests (stored in whatever format the client needs) and there are parameters for UI features, specifically extmarks, which are in vim's native UTF-8. IMO this is cleaner and more efficient than what we had before, where we ad-hoc converted from whatever client encoding to UTF-8.

rish987:
OK it looks like I've been contending with a bad decision I made way back in #113. I've reworked it to only ever take in UTF-8 params and auto-generate the client position params. Also removed the yucky extra `update()` parameters passed when setting position.

lua/lean/_util.lua
```diff
@@ -213,11 +213,33 @@ function M.position_params_valid(params)
...
   return true
 end
 
+-- REPLACEME: with vim.lsp.util.make_position_params(nil, "utf-8") after neovim/neovim#16382 is merged and
+-- we drop support for <0.6.0
+local function make_utf8_params()
+  local buf = vim.api.nvim_win_get_buf(0)
+  local row, col = unpack(vim.api.nvim_win_get_cursor(0))
+  row = row - 1
+  local line = vim.api.nvim_buf_get_lines(buf, row, row+1, true)[1]
```
gebner, line 227: The `vim.api.nvim_buf_line_count` function should be enough here, and doesn't require transfer of any buffer contents. (if it's necessary to check that the line is in range at all).

gebner (commented):
This is much better now.  AFAICT the `__ui_position_params` field is never sent via LSP, so there's no need to use LSP functions (with encodings) here at all, we could just use the UI rows/columns and pack them like `{ filename = ..., row = ..., col = ... }` (we also always convert the URI directly to a filename, so we might as well store the filename).

I was a bit surprised to see UTF-8 here, but I couldn't find any good documentation on what encoding the various neovim functions expect.  AFAICT we use the UI positions for extmarks and for display.
 1. Extmarks use UTF-8 positions, is that right?
 2. nvim_win_get_cursor is also UTF-8?
 3. For display, I think it would be best to show the window coordinates (which is also what airline does), but I think we could live with UTF-8.

lua/lean/_util.lua
rish987: Yea sorry this was dead code, removed.

rish987:
Okay reworked to avoid LSP params where possible. Yes vim and Lua are both natively UTF-8, you can assume it anywhere an encoding isn't mentioned.

lua/lean/infoview.lua
```diff
@@ -592,45 +596,64 @@ function Pin:remove_parent_info(info)
...
+  -- not a brand new extmark
+  if not buf then
+    buf = self.extmark_buf
+    local extmark_pos = vim.api.nvim_buf_get_extmark_by_id(buf, extmark_ns, self.extmark, {})
+    line = extmark_pos[1]
+    col = extmark_pos[2]
+  end
+
+  local buf_line = vim.api.nvim_buf_get_lines(buf, line, line + 1, false)[1]
+  local end_col = 0
```
gebner, line 633: I'm not sure I understand the logic here.  Is the code finding the closest valid UTF-8 position?  Does this actually happen (that we get an invalid UTF-8 position)?  If so, please explain this in a comment.

lua/lean/_util.lua
```diff
@@ -202,22 +202,38 @@ function M.list_workspace_folders()
...
 
   return true
 end
 
+function M.make_position_params()
+  local buf = vim.api.nvim_win_get_buf(0)
+  local row, col = unpack(vim.api.nvim_win_get_cursor(0))
+  row = row - 1
+
+  return { filename = vim.uri_to_fname(vim.uri_from_bufnr(buf)), row = row, col = col }
```
gebner, line 234:
```suggestion
  return { filename = vim.api.nvim_buf_get_name(buf), row = row, col = col }
```

lua/lean/infoview.lua
rish987: Added a comment. This gets the UTF-8 index of the next codepoint.

rish987:
Anything else? I think I'll merge this soon since it's backwards compatible anyways and seems like an all-around improvement. Will also mark the new tests as pending until neovim/neovim#16382 is merged.

gebner:
The test fails on nightly now.

rish987:
The tests have always been failing, as we're waiting on neovim/neovim#16382. I've locally patched it for our own requests but this is misleading since native LSP requests (hover, goto definition, etc.) will still be broken in 0.6. I could eventually monkey together our own custom LSP commands for Lean 3, but hoping this isn't necessary if neovim/neovim#16382 gets into 0.6.1.

PR #209, opened 2021-11-28
Put pins in separate windows.

rish987:
@Julian I'm coming around on your idea to put pins in their own window, though I don't think _each_ pin should have its own window, since this can get cluttered too quickly I think. This reflects a more general intent to treat the current pin as "sacred" in that it's the only thing you'll see in its window, which for example will help clean up state diffs if you have pinned something.


lua/lean/infoview.lua
```diff
@@ -236,12 +348,18 @@ function Infoview:close()
   end
 
   self:__close_diff()
+  self:__close_pins()
+
+  if vim.api.nvim_win_is_valid(self.pins_win) then
```
rish987, line 353: (note to self: should be `self.window`)

Julian:
I'll make some time to look at this in the next few days though I'm prepping for even more than usual this week so will be slow, apologies.

To clarify though -- I don't like pins because I find they complicate the implementation a lot, and provide functionality better served by native vim features. This opinion is of course uninformed, because I don't use them, so I could be talking complete nonsense, but that's where my previous comments have come from.  I bring that up again now because if you're coming around to that, this PR *adds* lines rather than removes, which worries me. At the minute infoview.lua tops 1000 lines (albeit with long emmy comments which bloat), but it just doesn't feel to me like it should be that large.

I'll try to give this and the other PR an honest objective shot nonetheless, and yeah in case I don't say it enough really do appreciate all of this work you've done!

gebner:
> To clarify though -- I don't like pins because I find they complicate the implementation a lot, and provide functionality better served by native vim features.

There are several features here, and only some of them can be replaced by vim features.  The original pin feature was that you can "pin" a position---instead of tracking the cursor position in the editor, the "pinned" info always shows information for the same "pinned" position.  Then you can also pause the info, so that it doesn't update when you change the file.  These clearly can't be done with native vim features.

The other part is that multiple pins can be shown in the same infoview.  I'm somewhat happy about this because I'm not so good at vim window management.  (But I don't get the extra `Info` layer in between `Infoview` and `Pin[]` either.)

But I think I could live with a different design where Infoview=Info=Pin and you had `d`, `p` and `u` keybindings to pin and disable updates, resp. (and also duplicate the infoview if it's attached to multiple windows).  Then the equivalent of `LeanInfoviewAddPin` would be `:spl<CR>p`; and there would finally be an obvious way to clear individual pins, namely `:q`!

gebner:
In case it's not clear: having all pins in a second window is the worst of both possible approaches for me.  I need to bother with both vim window management as well as the (limited and lean-specific) commands to manage pins.

rish987:
> But I don't get the extra Info layer in between Infoview and Pin[] either.

Does #211 make this clear? This gives us control of whether pins are shared between tabs.

> But I think I could live with a different design where Infoview=Info=Pin and you had d, p and u keybindings...

Not sure I'm understanding this part. Are you saying each pin should have its own split window? I feel like you would run out of space fast if you attempted this. BTW there actually is no way to clear individual pins currently, but it should be very easy to add a Div event for this.

> having all pins in a second window is the worst of both possible approaches for me.

I see, perhaps we can allow for the previous way of doing things (i.e. all pins in the same buffer) by having a "compact" mode, but if you're serious about using Lean you should probably have a large display anyways.

The main way I'm planning to use this is to `#print` a theorem and pin this so it shows in a separate window and I can step through the tactics in the written proof, and correlate things without my view on the pin constantly jumping around as the current pin changes. This also enables a clean state diff.

Julian:
I definitely don't think vim *fully* contains all the functionality we want, but I think you managed to lay out at least a concrete example of what I've been trying to say more clearly than I did:

>  Then the equivalent of LeanInfoviewAddPin would be :spl<CR>p; and there would finally be an obvious way to clear individual pins, namely :q!

This is precisely the kind of example I was trying to get across -- the existence already of ways to create windows-to-the-same-thing, and close those windows, and yes 100% agree that the missing piece we'd still need to add is related to pausing updates to them. (Actually there's maybe one additional thing which is that splits will be to the same buffer, so we also would need to teach the update code to notice when it has multiple windows to the same buffer with one paused how to handle that, but I believe this would be trivial to do).

> The other part is that multiple pins can be shown in the same infoview. I'm somewhat happy about this because I'm not so good at vim window management.

Fair enough, I have to say that in the opposite boat, the few times I've opened a pin to try them out for a PR review or something, it makes me have the reverse experience -- I *am* good at vim window management, but instead of being able to use what I know, I have to figure out some new way to do things, and worry we're gonna add a whole bunch of unneeded code to support these new ways. If I added 5 pins and I want to reorder them in how they display, are we going to add a bunch of code to do that? Because in vim, reordering windows is an existing thing with code we don't get to support.

rish987:
> If I added 5 pins and I want to reorder them in how they display, are we going to add a bunch of code to do that? Because in vim, reordering windows is an existing thing with code we don't get to support.

I guess it's a matter of _which_ upstream features we want to support. Because in doing this we no longer have the ability to scroll as you normally would in a vim buffer, and we also inherently limit how many pins we can have without things becoming unreadable.

gebner:
> I guess it's a matter of which upstream features we want to support. Because in doing this we no longer have the ability to scroll as you normally would in a vim buffer, and we also inherently limit how many pins we can have without things becoming unreadable.

How many pins do you typically have?  If each pin is a different window, you can also tile them so that they take up less space.

In an earlier comment you mentioned that scrolling is bad because then the pins jump around as you move through the file, or as the pins are being updated.  I've also found this annoying, in vscode as well.  In my eyes, this alone would be reason enough to use one window per pin.

rish987:
^ actually this may be more trouble than it's worth at the moment. I'll try to see if I can get per-pin windows to work in this PR. I typically only have one or two pins so yeah there shouldn't be much issue with screen space, and we can also eventually find some way to have a "pin-switcher" so we can easily swap out pins in the window without having to delete/re-add them.

rish987:
Okay @gebner @Julian this now creates one window per pin, also added some detailed tests.

rish987:
Interop should _almost_ be there in this PR, though we have a similar issue to #173 for the windows not recognizing when their buffer is changed. Plan to fix this in a subsequent PR (with tests, of course) so as to not overwhelm this one.

rish987:
One more thing I think we'll want to do is distinguish between `:close` and `:quit` so that `:quit` removes the pin from the info, whereas `:close` just closes the window and leaves the pin associated with the info. Then, we could `:close` a pin and then open its buffer in a new tab without the buffer being deleted. We could also have telescope integration to select which `:close`'d pin to replace in the pin window. Again, something for a later PR.

Julian:
We're still quite far away from each other designwise I think. I'm still hoping pins disappear entirely, and I don't see any benefit in the kind of functionality you're talking about with `:close` personally, or with say adding new ways to jump to windows, which I see you added now, rather than saying to go use a plugin to jump to windows if you want a better way to pick which window you want to go to. I still expect we'd be able to delete lots of lines if pins were just windows, which just distracts me when you talk about intricate pin functionality that personally I doubt others will use. I still don't have concrete thoughts though, I have to understand what this PR is trying to do and not do.

rish987:
I absolutely need pins. As far as I understand, you'd like to just copy the infoview text into a new buffer and split. Clearly it's not that simple -- what if I have printed something and want to `set_option pp.implicit true`? I have to go back to that position and copy it over again? I'm not trying to make this like an IDE, this just seems like the minimum amount of new code for allowing someone to use it with interactive theorem proving on a basic level.

Julian:
> I have to go back to that position and copy it over again?

If I understand the example properly, no that should work functionally the same as now -- I'm not saying we don't need ways to have infoviews pause (not automatically update, just sit there dead) or unpause (as normal), and also have them track the cursor position (normal) or some other position (what a pin does now).

> this just seems like the minimum amount of new code for allowing someone to use vim with interactive theorem proving on a basic level.

I think you're drastically overstating this functionality -- I don't think even you think this is basic functionality (pins, let alone each of the incremental pieces of functionality from these PRs). I have only anecdata, but from asking some VSCode users whether they know what pins are, no one said "I use them and they're essential", they either said they hadn't heard of them or use them rarely. Again though -- I'm not saying we shouldn't support the functionality at all though, just that the incremental complexity they add should match, and the design should fit within vim's norms, and I think we can improve on both before adding more functionality on top.

All that being said though, I haven't played with this PR, but if it does what it says on the tin and moves us to separate windows per pin, that's probably a nice improvement and we can merge and simplify later perhaps, so gimme a bit to play with it.

rish987:
> I have only anecdata, but from asking some VSCode users whether they know what pins are, no one said "I use them and they're essential", they either said they hadn't heard of them or use them rarely.

I want this to be usable for people interested in Lean's type-theoretic foundations. Sure if you don't use Lean beyond tactics you might be able to get away without this. So granted, it's not "basic", but certainly essential. Another bias here is that pins are _literally unusable_ in VSCode in their current state beyond trivial applications.

> the incremental complexity they add should match, and the design should fit within vim's norms, and I think we can improve on both before adding more functionality on top.

Again, I'm not sure what you mean by "complexity," I'd like to know if there's anything specific, like a function, conditional etc. that you think is more complex than necessary. Having said that, I have been noticing a few such areas while reworking autocmds a bit, so you may want to hold off on reviewing the code until tomorrow. While I'm at it, I think I'll just go ahead and fix #173 and implement what I was talking about with `:close` so that something like `<C-w> T` works as expected for pin windows. The goal with this PR is 100% interop :)

Julian:
> While I'm at it, I think I'll just go ahead and fix #173 

Please do *not* fix multiple things in one PR. This PR already does too much, PRs need to do one and only one thing. Simplify whatever is a simplification, but do not add additional fixes or features to one PR.

>  I'd like to know if there's anything specific, like a function, conditional etc. that you think is more complex than necessary.

I still believe the `Pin` object should not exist, and that one or both of `Infoview` and `Info` should just be `Window` and `Buffer`, and that there shouldn't be code that deals with opening and closing infoviews at different positions. That's the complexity I'm referring to -- the re-creation of objects (behavior) that already exist in vim with different names and slightly different behavior, that someone gets to learn. But I'm still not expressing that clearly it seems, even if somewhere Gabriel gave a concrete example that was spot on with part of what I meant, so I'm hoping to try to spend some time just showing what an alternate implementation looks like.

rish987:
> Please do not fix multiple things in one PR. This PR already does too much, PRs need to do one and only one thing. Simplify whatever is a simplification, but do not add additional fixes or features to one PR.

Got it, will branch.

> That's the complexity I'm referring to -- the re-creation of objects (behavior) that already exist in vim with different names and slightly different behavior, that someone gets to learn.

No one has to learn anything new. IMO all of our objects should even be private. Let me explain:
- A `Pin` is like an extra cursor, so you don't have to go back to a position if there's a state you want to keep visible, or an elaborated theorem you have printed for example. Each pin corresponds to a buffer.
- An `Info` is just a set of pins, it can have a current pin, diff pin, and extra pins.
- The `Infoview` is simply a set of windows that give you a "view" on an info. Each of its pins has its own window. If you `:quit` a pin's window it will be deleted, and if you `:close` it it will be closed but not deleted, so for example if you are sharing it with another infoview (i.e. tab) as in #211, its window there will not be closed. The infoview opens windows in an arrangement you'd probably want, but you can always move the windows around as you please.

These abstractions exist only for our own sake. Perhaps it would be more productive to discuss which aspects of the above you don't agree with?

gebner:
I've just tried out the PR.  To my surprise the pin was split vertically, i.e. the layout is:
```
 ------------------ 
|      |     |     |
| lean | pin | pin |
|      |     |     |
 ------------------ 
```
I wanted this instead:
```
 ------------------ 
|            | pin |
| lean        -----
|            | pin |
 ------------------ 
```

So I tried to do `ctrl-w l`, `:spl`, `:b lean://pin/123`, and `ctrl-w l :q` to get rid of the extra window.  Unfortunately this broke the infoview: it didn't update anymore.  Toggling the infoview just restores the vertical split that I don't want.  (I think I've mentioned this before but I'm still learning how to manage windows in vim.  I've since figured out that `ctrl-w l ctrl-w K ctrl-w j ctrl-w H` rearranges the windows without breaking the infoview.)

But it would still be great if `:q` didn't kill the infoview (if anything I would have expected it to remove that pin).

rish987:
> Unfortunately this broke the infoview: it didn't update anymore.

I can't reproduce this, so I suspect my steps are wrong. Which window did you `:q`, the current pin or the extra pin? By the sounds of it you quit the current pin (the one without the `-- ...` header) and were expecting the extra pin to update. Did updates work again when you re-opened? Also `:q` does mean remove that pin for an extra pin, but it means close the infoview for the current pin.

gebner:
> > Unfortunately this broke the infoview: it didn't update anymore.
> 
> I can't reproduce this, so I suspect my steps are wrong. Which window did you `:q`, the current pin or the extra pin? By the sounds of it you quit the current pin (the one without the `-- ...` header) and were expecting the extra pin to update. Did updates work again when you re-opened?

This is maybe an easier repro: do `:spl` and then `:q` in the current pin window.  Then you have a current pin that doesn't update.  Yes, closing and reopening worked.

> Also `:q` does mean remove that pin for an extra pin, but it means close the infoview for the current pin.

I find that a bit confusing.  As Julian says, buffers and windows should behave like normal buffers and windows and not have hidden action-at-a-distance built-in.

rish987:
> > Also `:q` does mean remove that pin for an extra pin, but it means close the infoview for the current pin.
> 
> I find that a bit confusing. As Julian says, buffers and windows should behave like normal buffers and windows and not have hidden action-at-a-distance built-in.

This is a pretty old test case from `run_spec.lua`, it depends on whether you want `:q` to indicate that the infoview is closed so that reopening works as expected. For interop I plan to make `:close` work as you would like here.

Julian:
vim users do not in my experience use both `:close` and `:quit`, I'm pretty sure most everyone just uses `:quit`, so I don't think we should make the two do different things for infoview windows. I know what both do, but yeah I never use `:close`.

(I don't have a response yet for the rest of the above, though I still find it a problem that you're defining key concepts in terms of pins, functionality I think most users of the plugin will never use, but I recognize you use the term differently than I do, so I can write out how I would define the key concepts sometime later.)

rish987:
> vim users do not in my experience use both `:close` and `:quit`, I'm pretty sure most everyone just uses `:quit`, so I don't think we should make the two do different things for infoview windows. I know what both do, but yeah I never use `:close`.

The distinction is that `:close` is automatic and `:quit` is manual, so for example `WinClosed` (but not `QuitPre`) is triggered on `<C-w> T`. Also there is a mapping for `:close`, `<C-w> c`, so I find it hard to believe that vim users don't use `:close`. In any case do we have a solution to wanting `:q` to normally indicate that the infoview is closed so that `<LocalLeader> i` reopens as expected afterwards, but not in this particular case?

rish987:
> I don't know of other plugins where one needs to use one or the other to change behavior, even ones that open windows.

"even ones that open windows" is the limiting factor here. If a plugin doesn't need to open a window, why would it even consider this?

> As end-users, the two do the same thing other than some behavior with the last window opened, which is why I believe most users use `:q`, and I also believe most users type `:q` rather than using the ^W mappings (or if they do they use ^Wq)

The difference is clear to me, `:q` means close this window along with any buffer-related side effect that this incurs. For example it's common to have an autocmd on `QuitPre` that deletes the buffer. `:close` means just close this window with minimal side-effects.

Your model seems to conflate pins and infoviews. In general it sounds like you only want to have the "pin" layer and simply remove the concept of a set of pins (Info) or view on such a set (Infoview) entirely. Sure we could do that, but the whole point of  these layers is to make window creation/deletion automatic for Lean.

Also a diff pin is indeed nothing special, it's just another pin, there's nothing distinguishing it at the Pin layer. We just store it in a variable called `Info.diff_pin` on the `Info` level and happen to open it in a window suitable for diffs on the `Infoview` level.

Julian:
This PR now adds the close/quit functionality, functionality I mentioned I don't want to merge, as well as does so wrapped up in this same PR?

rish987:
Well it was just a few lines after I made some general improvements, and I assumed we wanted it now because of @gebner's comment? But I can revert those lines.

lua/lean/infoview.lua
```diff
@@ -965,25 +990,27 @@ local function get_iv_from_win(win)
...
+    iv:__pin_quit(id)
+    just_quit = false
+  else
+    iv:__pin_closed(id)
+  end
 end
 
 --- A pin's window was quit.
 --- Will be triggered via a `QuitPre` autocmd.
----@param id number @pin id
```
rish987: (had to change this because apparently `<afile>` isn't populated by `QuitPre`)

rish987: Derp, I guess we're expected to use `nvim_get_current_win()` here. Just fixed.

Julian:
I haven't forgotten about these PRs FWIW -- not ignoring them, will have a look in a few more days when the semester is over and I can breathe a bit.

rish987:
No worries, take it easy and no rush at all on this! I want to make sure we can all agree on the design before we commit to anything.

PR #210, opened 2021-11-29
Make current pin unfocusable.

rish987:
The use case I have in mind is being able to "unfocus" the current pin so you can move around in the source, do hover requests etc. without the info text changing. We could emulate this before by pausing the current pin, but this seems much more intentional, and also gives you an extmark and interactive info.

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -1153,6 +1153,19 @@ function infoview.toggle_focus_curr_pin()
   end
 end
 
+function infoview.goto_curr_pin()
+  if not is_lean_buffer() then return end
+
+  local iv = infoview.get_current_infoview()
+  if iv ~= nil then
+    local curr_pin = iv.info.pin
+    local params = curr_pin.__ui_position_params
+    local uri_bufnr = vim.fn.bufnr(params.filename)
+    vim.api.nvim_set_current_buf(uri_bufnr)
+    vim.api.nvim_win_set_cursor(0, { params.row + 1, params.col })
+  end
+end
+
 function infoview.enable_widgets()
   local iv = infoview.get_current_infoview()
   if iv ~= nil then iv.info.pin:set_widget(true) end
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -73,6 +73,7 @@ local Pin = {next_id = 1}
 ---@field bufdiv BufDiv
 ---@field diff_bufdiv BufDiv
 ---@field win_event_disable boolean
+---@field focus_pin boolean
 local Info = {}
 
 --- A "view" on an info (i.e. window).
@@ -97,6 +98,13 @@ local diff_pin_hl_group = "LeanNvimDiffPin"
 vim.highlight.create(diff_pin_hl_group, {
   cterm = 'underline',
   ctermbg = '7',
+  gui   = 'underline',
+}, true)
+
+local curr_pin_hl_group = "LeanNvimCurrPin"
+vim.highlight.create(curr_pin_hl_group, {
+  cterm = 'underline',
+  ctermbg = '39',
   gui   = 'underline',
 }, true)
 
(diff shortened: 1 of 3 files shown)


lua/lean/infoview.lua
```diff
@@ -1062,6 +1092,15 @@ function infoview.toggle_auto_diff_pin(clear)
   infoview.get_current_infoview().info:toggle_auto_diff_pin(clear)
 end
 
+function infoview.toggle_focus_curr_pin()
+  if not is_lean_buffer() then return end
+
+  local iv = infoview.get_current_infoview()
+  if iv ~= nil then
+    iv.info:toggle_focus_pin(vim.lsp.util.make_position_params())
```
rish987, line 1100: Note to self: remember to update after #208.

gebner:
> We could emulate this before by pausing the current pin, but this seems much more intentional, and also gives you an extmark and interactive info.

I don't think we should add multiple commands that do almost the same thing.  Adding an extmark for the current pin if it is paused seems reasonable enough.

rish987:
> I don't think we should add multiple commands that do almost the same thing. Adding an extmark for the current pin if it is paused seems reasonable enough.

It's not quite the same thing. This is a lightweight alternative to making a new pin, with all the benefits that come with pins being able to be dynamically updated. For example if I have a `#print ...` command and want to try out various `pp` options above, I can just unfocus the current pin at the `#print` and do so without having to clutter up the infoview without an extra pin.

gebner:
> For example if I have a #print ... command and want to try out various pp options above, I can just unfocus the current pin at the #print and do so without having to clutter up the infoview with an extra pin.

Then I completely misunderstood what unfocus does.  So it's not like pause at all, but like a "temporary" pin.  And you can simulate it with AddPin/ClearPins (at least if there are no other pins).  For composability it might be more useful to have AddPin add pins to the top, and add a ClearFirstPin command.  Then you could also unfocus twice.

rish987:
> And you can simulate it with AddPin/ClearPins (at least if there are no other pins). For composability it might be more useful to have AddPin add pins to the top, and add a ClearFirstPin command. Then you could also unfocus twice.

Not sure I'm understanding, there is always a current pin, so adding a pin would result in a second pin in the infoview, which is the clutter I want to avoid. Though indeed it may be useful to allow pins to be inserted at the top of the pin list rather than the bottom. What do you mean by "unfocus twice"?

gebner:
> Not sure I'm understanding, there is always a current pin, so adding a pin would result in a second pin in the infoview, which is the clutter I want to avoid.

Right, it won't do anything about the clutter.  (Though there is almost no clutter with the set_option example.)

> What do you mean by "unfocus twice"?

With unfocus, you can temporarily pin one position to see the effect of set_option.  But if you want to look at *two* positions, then you're out of luck.

I don't think that this is the most common case, but I think it is worth it to make general abstractions that can be combined in more than one way if possible.

rish987:
> With unfocus, you can temporarily pin one position to see the effect of set_option. But if you want to look at two positions, then you're out of luck.

In this case you'll have to place a new pin, since you actually _do_ want to see two things at once.

Another important use of this is state diffs, since these are always done on the current pin vs. the diff pin. If there's somewhere else in the proof you want to hover but you don't want to lose your diff, you can just unfocus and do whatever you want.

rish987:
Oops, didn't mean to close this one, will re-open as draft for now. I've found myself wanting this in particular for the `simp only` tactic, so I can leave my cursor at the end, try out different lemmas and see the effects. Certainly open to discussion and let me know if there's another obvious way I should be doing this.

Julian:
Hi! Cool. I still don't 100% follow this one honestly (not that I disagree with it, I literally don't understand which functionality it wants to implement) -- I've noticed in my own simple things that it's quite annoying when you're tinkering with something that refreshes the infoview that you lose your ability to read / think about what was there previously while it's updating (e.g. while waiting for a long `simp`). Is that what you mean? Or something else?

Basically yeah sorry can you describe or record what functionality this one is meant to implement without using new terminology ("focusing")?

PR #211, opened 2021-12-01
Allow to share infos between infoviews.

### LINKED ISSUE #126: Option to re-use info on a new tab
When opening a new tab, right now we always create a new info, but we should have an option that can enable using the same info in a new infoview. This should also come with a command to replace the current info with a brand new one if you actually want a new info.

rish987:
Closes #126. My recent workflow has been to have one tab with mathlib, and another tab with a leaf scratch Lean file where I print definitions, theorems, etc. It would be very useful for any pins in this scratch tab to be shared with the mathlib tab. Will add tests soon...


Julian:
Just to save you some possible effort, I'd probably recommend waiting on spending extra effort on tests -- like I mentioned in #209, I want to make sure each of these is worth the additional complexity before merging any of them, so you may want to hold off on spending extra effort until then, just in case we close this or any of the other ones.

Here for instance, I'm not sure I understand what this does, so a demo would help, but if I can guess, why can't you do this by just opening the same infoview buffer in your other tab (using the normal ways to do so in vim)? If that doesn't work because of some assumed 1-to-1 mapping between infoview windows and buffers, I'd prefer to fix that.

gebner:
> It would be very useful for any pins in this scratch tab to be shared with the mathlib tab.

I thought this just works if you add a pin?  If you set a pin in tab a, then it's still there if you switch to tab b.

rish987:
> why can't you do this by just opening the same infoview buffer in your other tab (using the normal ways to do so in vim)?

That is exactly what this does. We're using the same buffer in the second infoview. This just automates that process.

> If that doesn't work because of some assumed 1-to-1 mapping between infoview windows and buffers, I'd prefer to fix that.

There is no such assumption, hence why this works with very little changes.

> I thought this just works if you add a pin? If you set a pin in tab a, then it's still there if you switch to tab b.

 I don't see how this could work since we're currently creating new infos for every infoview. This works for you?

Julian:
> That is exactly what this does. We're using the same buffer in the second infoview. This just automates that process.

What I am saying is I don't want plugin-specific ways to do these things. What's wrong with the vim-native ones? In vim, opening an existing buffer in another window is `:b somebufname`. E.g. `:b info/1`. Or more typically these days, you use a fuzzy buffer-finding plugin of your choice, like telescope. Automating something like that is as simple as adding a mapping if it's something you do often.

What would be required for us to be able to do that, rather than adding specific extra code to infoview.lua?

rish987:
> What I am saying is I don't want plugin-specific ways to do these things. What's wrong with the vim-native ones?

Maybe I'm missing something, but isn't that the whole point of a plugin? To automate patterns common to your specific application?

> In vim, opening an existing buffer in another window is `:b somebufname`. E.g. `:b info/1`... What would be required for us to be able to do that

What I want here is the ability to also make pin in the new infoview and have it be added to that shared info. So what would be required here is adding new `BufEnter` autocmds to reset the Info when changing the buffer in the infoview. That would be interesting to do, but it should come on top of what we have here, because this gives a way to avoid having created a new info in the first place.

Julian:
> Maybe I'm missing something, but isn't that the whole point of a plugin? To automate patterns common to your specific application?

No -- just ones that are otherwise hard to build themselves, and that are likely to be used by some decent fraction of users of the plugin. Someone building a whole infoview setup is not tractable, it's been tricky work for us. That's the job of the plugin, to give enough richness to use as a foundation. Automating every possible thing someone may want to do with Lean + vim is not. It's just unmaintainable to do so.

Nor is it the job of the plugin replacing native functionality -- I want the plugin to play nicely with vim's functionality, not make it so it's impossible to use other plugins alongside it which rely on things working for lean.nvim the same as they do for any other window or buffer. Vim is not VSCode, and I don't want this plugin to be a Lean IDE within neovim that otherwise is totally isolated from the rest of the ecosystem because you can't use the same tools you use for other uses of vim.

> What I want here is the ability to also make pin in the new infoview and have it be added to that shared info. So what would be required for this is adding new BufEnter autocmds to reset the Info when changing the buffer in the infoview. 

I don't follow this yet but it doesn't sound like something that should be impossible to overcome. Can you elaborate on some short steps? It's "open lean file, create a pin, open new tab, open same lean file, create another pin from that other window to the same lean buffer, then go back to the first one and still see the pin shown in the first tab's infoview window"?

rish987:
> I want the plugin to play nicely with vim's functionality, not make it so it's impossible to use other plugins alongside it which rely on things working for lean.nvim the same as they do for any other window or buffer.

I think the only veritable bug in this area is #173, and this should just be a simple matter of registering the infoview as closed if you `BufLeave` one of its windows.

> Vim is not VSCode, and I don't want this plugin to be a Lean IDE within neovim that otherwise is totally isolated from the rest of the ecosystem because you can't use the same tools you use for other uses of vim.

I think we need to acknowledge that we are building something new to Vim here. Lean is by far the most interactive language server, and interactive theorem proving is a very specific use case that requires a lot of new code -- see Emacs' Proof General,  for example. We should eventually decouple the Infoview into its own plugin for general use by many different language servers.

> Can you elaborate on some short steps?

1. Open a lean file and place some pins.
2. Open a different lean file in a new tab, the same pins should be present in the new infoview.
3. Open a different lean file in a third tab and run some command so that you get a blank infoview (no pins) in this tab.

Julian:
> I think we need to acknowledge that we are building something new to Vim here. Lean is by far the most interactive language server, and interactive theorem proving is a very specific use case that requires a lot of new code

Sorry, we disagree here, which is why I'm pushing back on these PRs. I do not believe we are that different. I do believe there are certain things that are new, and that we need code for them. This does not strike me as one of them, nor do other PRs right now.

What I want instead is to reduce the amount of complexity we already have. It makes me uncomfortable, and I've already found it hard to work on the infoview code.

I'm not saying no new functionality is useful, which is why the PRs you have may very well be useful. But no, I don't agree with this general acknowledgement, and I'm not ready to merge a bunch of additional infoview PRs without careful consideration. I want to be clearer on that at the minute -- I've tried to say so previously that we're drifting too far away for my comfort. Hopefully being clearer now will save some extra wasted effort.

I was not referencing any specific bug. As I say, what I'm referencing is possible integration someone may want with other plugins that rely on things working with native vim concepts, and therefore with minimizing the amount of Lua API that does custom manipulation of internal state in vastly different ways. Such things are likely to cause integration issues with other plugins.

> We should eventually decouple the Infoview into its own plugin for general use by many different language servers.

This may well be a useful idea, and I opened https://github.com/whonore/Coqtail/issues/237#issuecomment-960025091 just to let the Coq folks know we exist, they very well may be interested.

rish987:
> What I want instead is to reduce the amount of complexity we already have. It makes me uncomfortable, and I've already found it hard to work on the infoview code.

To be honest I find the existing code quite easy to follow, and I don't think that's just because I'm the one who wrote a lot of it. I just see it as thin OOP wrappers on what vim natively provides. I would love to help and you can always ping me if you have issues, but I just don't know the specifics of what you're having trouble with.

gebner:
>> I thought this just works if you add a pin? If you set a pin in tab a, then it's still there if you switch to tab b.
>
> I don't see how this could work since we're currently creating new infos for every infoview. This works for you?

I only have a single infoview, but if I add a pin the pin just stays there when I go to another file.  I can even interact with the widgets from the other files (requires `se hid` though).

rish987:
> I only have a single infoview, but if I add a pin the pin just stays there when I go to another file. I can even interact with the widgets from the other files (requires `se hid` though).

I see, so you are buffer-switching, I am talking about multiple tabs since infoviews are created per-tab.

gebner:
Oh, I have to admit I've never used tabs before.  I'll check them out.

gebner:
> In vim, opening an existing buffer in another window is :b somebufname. E.g. :b info/1. Or more typically these days, you use a fuzzy buffer-finding plugin of your choice, like telescope.

Speaking of this, the infoview buffer does not show up as an autocompletion in `:b` for me.  I can still do `:b lean://info/1/curr`, but I can't find it with `:b lean<TAB>`.  Are we disabling this somehow?

Julian:
We're doing:

https://github.com/Julian/lean.nvim/blob/main/lua/lean/html.lua#L346

as I believe the line that ultimately creates the infoview buffer. The first `false` there sets `nolisted` on the buffer. Changing that to true seems reasonable to fix `:b` though I never use `:b` personally.

(It still also rubs me wrong that buffer creation for an infoview happens inside an object called `html.BufDiv`, but I'm pretty sure that's the line that's relevant.)

Julian:
Uh and on second thought because what's probably happening there is infoview windows and widget popups somehow being created in similar ways, probably that simple fix isn't enough, because having all the widget popups be listed buffers would possibly be annoying (though possibly not, it's an alternate navigation mechanism). So it may not be as simple as flicking false -> true. But yeah IMO we should be separating the two from each other if they're combined in that way.

gebner:
Flicking the flag was straightforward: https://github.com/Julian/lean.nvim/pull/213

For #212 I've wanted a BufDiv:new version that can take an existing buffer anyhow.

> It still also rubs me wrong that buffer creation for an infoview happens inside an object called html.BufDiv.

The naming is a bit misleading indeed, it has nothing to do with html or html's div element.  I'm not sure if there's a significantly better name though.  Would `widgets.Elem` and `widgets.BufRenderer` instead of `html.Div` and `html.BufDiv` be better?

Julian:
> Would widgets.Elem and widgets.BufRenderer instead of html.Div and html.BufDiv be better?

+1 both on those names as well as trying to make sure creating them doesn't have side effects.

rish987:
In retrospect on #213 I'm unsure why we're against `BufDiv` having the "side effect" of creating its own buffer. It already has to create buffers for its tooltips, so I didn't think it was a big leap for it to make its own buffer and be entirely self-contained. It seems we've just pushed this side effect onto `Info` instead, and of course the buffer has to be created on some level. Besides `BufDiv` literally has "buf" in the name, perhaps you were confusing it with the much more prolific `Div`?

Julian:
Besides the specifics of the example it sounded like #212 had, I don't know what norms Lua has, but normally elsewhere personally I would never expect creating an object to have side effects, those should happen in some method call. E.g. `Infoview:new` should not open a window either, that should happen by its caller calling `:open`, and that should happen in something other than the caller's `:new` which also should be side-effect free. These objects of ours are essentially impossible to unit test (vs. integration test, and probably that's fine), but normally you at least want it to be possible in theory to do so I think, and without being able to create objects you're kind of SOL.

rish987:
^ SGTM, I never thought of it that way. Should be easy to fix all of these side-effects, though this should be separate from the new features introduced in these PRs.

Julian:
Yes 100% agree on it being some unrelated thing, not even a super important one to me just elaborating on why I agreed with doing it once there was another good reason to do so.

(Normally I'm a believer in something Michael Feathers claims, which is, to paraphrase "anything that makes something harder to test will also have a *non-testing* related thing you'll care about that it makes harder too", so Gabriel's example just reminded me that I'd noticed this before)

PR #212, opened 2021-12-02
Add :LeanGoal and :LeanTermGoal commands for interactive popups. 

gebner:
This is an interactive upgrade to the `:LeanPlainGoal` and `:LeanTermGoal` commands.

### DIFF
--- a/lua/lean/commands.lua
+++ b/lua/lean/commands.lua
@@ -15,7 +15,6 @@ local function show_popup(div)
   local bufnr, winnr = vim.lsp.util.open_floating_preview(
     vim.split(div:to_string(), '\n'), 'leaninfo',
     { close_events = {} })
-  vim.api.nvim_buf_set_option(bufnr, "bufhidden", "wipe")
 
   local bufdiv = html.BufDiv:new(bufnr, div, infoview.mappings)
   bufdiv.last_win = winnr
--- a/lua/lean/commands.lua
+++ b/lua/lean/commands.lua
@@ -13,7 +13,7 @@ local plain_term_goal = a.wrap(leanlsp.plain_term_goal, 3)
 ---@param div Div
 local function show_popup(div)
   local str = div:to_string()
-  if str:match('%s*') then
+  if str:match('^%s*$') then
     -- do not show the popup if it's the empty string
     return
   end
--- a/lua/lean/commands.lua
+++ b/lua/lean/commands.lua
@@ -12,8 +12,14 @@ local plain_term_goal = a.wrap(leanlsp.plain_term_goal, 3)
 
 ---@param div Div
 local function show_popup(div)
+  local str = div:to_string()
+  if str:match('%s*') then
+    -- do not show the popup if it's the empty string
+    return
+  end
+
   local bufnr, winnr = vim.lsp.util.open_floating_preview(
-    vim.split(div:to_string(), '\n'), 'leaninfo',
+    vim.split(str, '\n'), 'leaninfo',
     { focus_id = 'lean_goal' })
 
   local bufdiv = html.BufDiv:new(bufnr, div, infoview.mappings)
--- a/lua/lean/commands.lua
+++ b/lua/lean/commands.lua
@@ -14,7 +14,7 @@ local plain_term_goal = a.wrap(leanlsp.plain_term_goal, 3)
 local function show_popup(div)
   local bufnr, winnr = vim.lsp.util.open_floating_preview(
     vim.split(div:to_string(), '\n'), 'leaninfo',
-    { close_events = {} })
+    { focus_id = 'lean_goal' })
 
   local bufdiv = html.BufDiv:new(bufnr, div, infoview.mappings)
   bufdiv.last_win = winnr
--- a/lua/lean/commands.lua
+++ b/lua/lean/commands.lua
@@ -1,0 +1,80 @@
+local components = require'lean.infoview.components'
+local infoview = require'lean.infoview'
+local leanlsp = require'lean.lsp'
+local a = require'plenary.async'
+local html = require'lean.html'
+local rpc = require'lean.rpc'
+
+local commands = {}
+
+local plain_goal = a.wrap(leanlsp.plain_goal, 3)
+local plain_term_goal = a.wrap(leanlsp.plain_term_goal, 3)
+
+---@param div Div
+local function show_popup(div)
+  local bufnr, winnr = vim.lsp.util.open_floating_preview(
+    vim.split(div:to_string(), '\n'), 'leaninfo',
+    { close_events = {} })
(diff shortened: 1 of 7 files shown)


gebner:
I haven't found a good way to enter the tooltip so far.  ctrl-w ctrl-w enters the infoview instead..

Julian:
I get some errors here when trying to run these commands on a spot with no goal info (ultimately saying `open_floating_popup` got called with a width of 0 I think) -- not sure I see why yet from reading the code but I didn't try to debug yet, will give it some more time later.

Julian:
Also some weird behavior that the popup isn't disappearing for me when I move away... not sure why yet either.

Julian:
> I haven't found a good way to enter the tooltip so far. ctrl-w ctrl-w enters the infoview instead..

Normally this is I think by focusing the same float ID again.. (from :h vim.lsp.util.open_floating_preview()):

>                                 • focus: (boolean, default true) If `true` ,
>                                  and if {focusable} is also `true` , focus an
>                                  existing floating window with the same
>                                  {focus_id}

So you should set `{ focus_id = "lean_goal"}` say, and then if someone uses `:LeanGoal` twice, they'll jump into the floating window. I actually really hate this behavior normally (e.g. in `.hover()`) so I turn it off everywhere, but it's the idiom people seem to be using right now, so it's at least an option.

lua/lean/html.lua
```diff
@@ -357,7 +357,8 @@ function BufDiv:new(buf, div, keymaps)
   util.set_augroup("DivPosition", string.format([[
     autocmd CursorMoved <buffer=%d> lua require'lean.html'._by_id[%d]:buf_update_cursor()
     autocmd BufEnter <buffer=%d> lua require'lean.html'._by_id[%d]:buf_update_cursor()
-  ]], buf, id, buf, id), buf)
+    autocmd BufDelete <buffer=%d> lua require'lean.html'._by_id[%d] = nil
```
rish987, line 360: From what I can tell this has no effect, that is when I open a tooltip and close it repeatedly and do

```
:lua for id, _ in pairs(require'lean.html'._by_id) do print(vim.inspect(id)) end
```

that list only ever grows, which makes me suspect refs aren't getting released as they should. Not sure what's going on, perhaps we need a different autocmd.

rish987: Yes I've confirmed that refs are no longer getting released at all if the `BufDiv` corresponding to a tooltip survives. In general this seems like a more fragile way of doing things, could we instead make a local `BufDiv` in `commands.lua` that is `nil`'ed when the popup is closed?

gebner:
> I get some errors here when trying to run these commands on a spot with no goal info (ultimately saying open_floating_popup got called with a width of 0 I think) -- not sure I see why yet from reading the code but I didn't try to debug yet, will give it some more time later.

I didn't notice this because it doesn't happen in Lean 4.  Your theory was correct, `open_floating_popup` was called with a whitespace-only string.  I've put in a check for that.

> Also some weird behavior that the popup isn't disappearing for me when I move away... not sure why yet either.

That was (semi-)intentional by setting `close_events = {}`.  It seemed useful to show the goal while you're still editing the tactic.  I'll change it back to the default if you don't like it.

> So you should set { focus_id = "lean_goal"} say, and then if someone uses :LeanGoal twice, they'll jump into the floating window. I actually really hate this behavior normally (e.g. in .hover()) so I turn it off everywhere, but it's the idiom people seem to be using right now, so it's at least an option.

Thank you for the explanation!  I've noticed that sometimes hovering enters tooltip, but I could never figure out why.  For the interactive hovers it seems kind of useful though, so I've put in the `focus_id`.

lua/lean/html.lua
gebner: Good catch. Okay, I've switched back to the weak table and put the `bufdiv` variable in `commands.lua`.  The downside is that you can't have tooltips open in multiple windows at the same time.

> nil'ed when the popup is closed?

Unfortunately, I couldn't find any reliable way to either detect when the floating window is closed or when the buffer is deleted.  Autocmds like `BufDelete`, `BufUnload`, `WinClose` don't seem to trigger for this popup???  Weirdly enough, `BufUnload` triggers for the tooltips in the infoview (but not `BufDelete`).

If we could reliably detect when a buffer is deleted, then I think this would be the cleanest solution here.  You wouldn't need to keep a reference to the `BufDiv` object (which is awkward for popups), and it doesn't introduce any leaks either (it only leaks the BufDiv if the buffer is leaked).

lua/lean/commands.lua
```diff
@@ -0,0 +1,88 @@
...
+local html = require'lean.html'
+local rpc = require'lean.rpc'
+
+local commands = {}
+
+local plain_goal = a.wrap(leanlsp.plain_goal, 3)
+local plain_term_goal = a.wrap(leanlsp.plain_term_goal, 3)
+
+-- Keep reference around to prevent garbage collection
+local bufdiv
```
rish987, line 14: Could we instead use a mapping from popup window handle to bufdiv that is `nil`ed on `WinClosed`? You can use `vim.fn.expand('<afile>')` to get the window handle in the autocmd.

rish987: Oh just saw your comment above, `WinClosed` is not working for the popup window?

rish987: Hmm so it triggers when I `:q` from within the window but not when it closes automatically after I move the cursor...

gebner: I tried `autocmd WinClosed * echo "wont be printed"`, and it does not seem to be working for the floating popups

lua/lean/html.lua
rish987: > If we could reliably detect when a buffer is deleted, then I think this would be the cleanest solution here.

I agree, but it looks like we're running into upstream issues, though I haven't found any existing reports on this. So I guess we'll have to deal with it for now.

rish987:
LGTM at this point, I would also like to persist `BufDiv`s along with their buffers, but we may have to file some issues upstream relating to these broken autocmds.

lua/lean/commands.lua
rish987: Could we perhaps work around this at the moment by checking `vim.api.nvim_win_is_valid()` for each handle in this table and `nil`ing accordingly whenever a new popup is created?

gebner: If we go that route, then I'd rather check for `vim.api.nvim_buf_is_valid` so that we can make `_by_buf` non-weak.

gebner: I've implemented that now.

rish987: Great, that seems like a good solution for now.

gebner:
> But that's just a minor annoyance -- if you have a second wouldn't mind if that just silently did nothing on Lean3

Okay, I've put in a check.

PR #213, opened 2021-12-02
List infoview buffer in :b

gebner:
Conflicts with #212 (which also wants a BufDiv:new function which takes an existing buffer).


gebner:
Disturbingly, I have just noticed that the diff features will always render the diff buffer even the diff is not enabled? :-/

rish987:
> Disturbingly, I have just noticed that the diff features will always render the diff buffer even the diff is not enabled? :-/

Yes my bad, just fixed.

rish987 (approved):

PR #214, opened 2021-12-02
Allow for unregistering `BufDiv` buffer.

rish987:
This should improve interop as suggested by Julian. Specifically it will disable autocmds and mappings in the (unlisted) diff buffer if it's not open, so that if someone manages to find it it will just be like a normal buffer there won't be any errors related to the `nil` div.


rish987:
I'm rethinking this ahead of an anticipated rework of #209 wherein each pin (including the diff pin) has its own BufDiv. So ignore this for now.

rish987:
Closing in favor of #209.

PR #216, opened 2021-12-07
Add LeanLineDiagnostics command.

gebner:
Interactive version of `vim.lsp.diagnostic.show_line_diagnostics`.

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -36,6 +36,7 @@ local options = {
     indicators = "auto",
     lean3 = { show_filter = true, mouse_events = false },
     show_processing = true,
+    show_no_info_message = false,
     use_widget = true,
 
     mappings = {
@@ -765,6 +766,7 @@ function Pin:__update(tick, delay, lean3_opts)
   delay = delay or 100
 
   self:set_loading(true)
+  local blocks = {} ---@type Div[]
   local new_data_div = html.Div:new("", "pin-data", nil)
 
   if delay > 0 then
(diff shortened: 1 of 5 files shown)


lua/lean/_util.lua
```diff
@@ -5,6 +5,15 @@ local a = require("plenary.async")
 
 local M = {}
 
+---@generic T
+---@param dest T[]
+---@param src T[]
+function M.insert_all(dest, src)
```
Julian, line 11: Isn't this vim.list_extend?

lua/lean/infoview.lua
```diff
@@ -36,6 +36,7 @@ local options = {
     indicators = "auto",
     lean3 = { show_filter = true, mouse_events = false },
     show_processing = true,
+    show_no_info = false,
```
Julian, line 39:
```suggestion
    show_no_info_message = false,
```

Bit clearer to me perhaps what this does, even if it's longer.

lua/lean/infoview.lua
```diff
@@ -842,16 +844,10 @@ function Pin:__update(tick, delay, lean3_opts)
...
-    if not goal_div_empty and not term_goal_div_empty then
-      new_data_div:add_div(html.Div:new("\n\n", "plain_goal-term_goal-separator"))
-    end
-    new_data_div:add_div(term_goal_div)
-
-    if goal_div_empty and term_goal_div_empty then
-      new_data_div:add_div(html.Div:new("No info.", "no-tactic-term"))
+    util.insert_all(blocks, goal_div)
+    util.insert_all(blocks, term_goal_div)
+    if options.show_no_info and #goal_div + #term_goal_div == 0 then
```
Julian, line 849:
```suggestion
    if options.show_no_info_message and #goal_div + #term_goal_div == 0 then
```

Julian (approved):
lgtm other than the minor comments

lua/lean/_util.lua
gebner: Yes, you're right.  I usually try to avoid the `vim.*` functions because they have lots of performance issues (see e.g. https://github.com/neovim/neovim/pull/16197 where list_extend was removed for that reason).  But this is not a performance critical part, so let's use list_extend.

PR #218, opened 2021-12-13
Use vim.lsp.client instead of buf_request.

### LINKED ISSUE #177: LSP method not supported errors on startup
I often see messages like:

```
method $/lean/discoverWidget is not supported by any of the servers registered for the current buffer
```

when starting up nvim. Presumably if requests are being sent to the LSP before it's fully ready?

It will happen with similar messages mentioning one or two other language servers methods as well. Everything else works fine, and the infoview happily works as I start to edit.

gebner:
Fixes #177.

### DIFF
--- a/lua/lean/lean3.lua
+++ b/lua/lean/lean3.lua
@@ -57,6 +57,9 @@ local to_event = {
 
 function lean3.update_infoview(pin, data_div, bufnr, params, use_widget,
     opts, options, show_processing, show_no_info_message)
+  local client = lsp.get_lean3_server(bufnr)
+  if not client then return end
+
   local parent_div = html.Div:new("", "lean-3-widget")
   local widget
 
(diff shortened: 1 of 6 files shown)


lua/lean/lean3.lua
```diff
@@ -57,6 +57,9 @@ local to_event = {
 
 function lean3.update_infoview(pin, data_div, bufnr, params, use_widget,
     opts, options, show_processing, show_no_info_message)
+  local client = lsp.get_lean3_server()
```
rish987, line 60:
```suggestion
  local client = lsp.get_lean3_server(bufnr)
```
necessary for pins

lua/lean/lsp.lua
```diff
@@ -14,46 +12,51 @@ function lsp.enable(opts)
...
+--- given bufnr.
+function lsp.get_lean3_server(bufnr)
+  local lean_client
+  vim.lsp.for_each_buffer_client(bufnr, function (client)
+    if client.name == 'lean3ls' then lean_client = client end
+  end)
+  return lean_client
+end
+
 -- Fetch goal state information from the server.
```
rish987, line 35: I think it would be a good idea to mention in the header when functions are async.

gebner: I've added it to the docstring.  In general, all our LSP functions are async though.

PR #219, opened 2021-12-13
Disable incremental sync in neovim 0.6.0

### LINKED ISSUE #217: Inserting a space before ⟨ crashes the server
I'm not sure if this is a bug in the lean 4 server or in nvim 0.6.

To reproduce, create a file with the single character `⟨`.  Then insert a space before.  You should see this error:
```
Watchdog error: Cannot read LSP message: offset 119: unexpected character in object
```

The lean 3 server seems to be fine.  Do you see the same error?

gebner:
Workaround for https://github.com/neovim/neovim/issues/16624

Fixes #217

### DIFF
--- a/lua/lean/lsp.lua
+++ b/lua/lean/lsp.lua
@@ -11,6 +11,10 @@ function lsp.enable(opts)
       vim.lsp.handlers['textDocument/publishDiagnostics'](...)
     end;
   })
+  if vim.version().major == 0 and vim.version().minor >= 6 then
+    -- workaround for https://github.com/neovim/neovim/issues/16624
+    opts.flags = vim.tbl_extend('keep', opts.flags or {}, { allow_incremental_sync = false })
+  end
   require('lspconfig').leanls.setup(opts)
 end
 


gebner:
I couldn't reproduce the sync error with Lean 3, so disabling it only for Lean 4.

mjlbach:
FYI the multibyte handling is *completely* incorrect in 0.5.0 and 0.5.1, so I would disable there too.

gebner:
Thanks for the heads up!  I haven't seen any issues before, so maybe we've just been lucky.  I've now disabled incremental sync on all neovim versions.

mjlbach:
Try adding a multibyte character (like `你好 再见`) and delete/undo `好 再`

gebner:
> Try adding a multibyte character (like 你好 再见) and delete/undo 好 再

That works for me in 0.5.1.

gebner:
I think I was too fast with disabling this for Lean 4, the Lean 4 server is broken with full synchronization.  It always reprocesses the entire file now...  We probably need to revert the workaround..

mjlbach:
It won’t throw an error but the buffer state server side will be wrong
________________________________

gebner:
I'm sure there were plenty of unicode bugs, but this doesn't seem to be it.  I can delete any character from the following, undo it, and everything's fine again.

```lean
def «你好» := 1
def «再见» := 2

#check «你好» + «再见»
```

Julian:
> I think I was too fast with disabling this for Lean 4, the Lean 4 server is broken with full synchronization. It always reprocesses the entire file now... We probably need to revert the workaround..

Done (in 1837371196e20380dfb348f38938c8bc232ef5d7), but I didn't test personally.

PR #227, opened 2022-01-31
Some performance tweaks for widgets

gebner:
The infoview has gotten noticably slower with large trace messages since last year.  It now seems roughly back to the state in December.  (No hard numbers, sorry.)

### DIFF
--- a/lua/lean/widgets.lua
+++ b/lua/lean/widgets.lua
@@ -7,16 +7,13 @@ local util = require('lean._util')
 ---@field events table<string, fun()> @event function map
 ---@field text string @the text to show when rendering this element
 ---@field name string @a named handle for this element, used when path-searching
----@field hlgroup string|fun():string @the highlight group for this element's text, or a function that returns it
+---@field hlgroup? string|fun():string @the highlight group for this element's text, or a function that returns it
 ---@field tooltip? Element Optional tooltip
 ---@field highlightable boolean @(for buffer rendering) whether to highlight this element when hovering over it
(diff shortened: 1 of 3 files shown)


lua/lean/widgets.lua
```diff
@@ -37,20 +34,33 @@ local BufRenderer = {
...
+
 ---Create a new Element.
+---@param args ElementNewArgs
 ---@return Element
-function Element:new(obj)
-  obj = obj or {}
-  local children = obj.children or {}
-  obj.children = nil
-  return setmetatable(
-    vim.tbl_extend("keep", obj, { events = {}, __children = children }),
```
gebner, line 47: `tbl_extend` was 12% of the runtime btw.

Julian: Eesh.

Julian:
Thanks!

Julian:
Is there a specific case you were performance optimizing with by the way? Or just something similar to what you've shared before that spits out 1000 trace messages? Just want to make sure I know what to check in the future.

gebner:
Yes, it's just the one with the adjustable number of trace messages.  I should probably also look at a benchmark with collapsible traces.

```lean
def Good (n : Nat) := True

@[simp] theorem Good_zero : Good 0 := ⟨⟩
@[simp] theorem Good_succ_iff : Good (n+1) ↔ Good n := Iff.rfl

set_option trace.Meta.Tactic.simp true in
example : Good 5000 := by simp
```

PR #228, opened 2022-01-31
Get rid of `lean3_opts` in infoview.

gebner:
The main goal of this PR is to remove the circular infoview <-> Lean 3 widgets dependency, where Lean 3 widgets send events to the server by updating the infoview with a special `lean3_opts` argument.  This made any changes to the updating logic annoying, because you always had to keep track of this special code path for Lean 3.

There's now a total of six lines in `infoview.lua` for Lean 3 support.

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -704,7 +704,7 @@ function Pin:__finished_loading()
   return true
 end
 
-function Pin:async_update(force, _, lean3_opts)
+function Pin:async_update(force, lean3_opts)
   if not force and self.paused then return end
 
   local tick = self.__ticker:lock()
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -721,113 +721,109 @@ end
 
 Pin.update = a.void(Pin.async_update)
 
---- async function to update this pin's contents given the current position.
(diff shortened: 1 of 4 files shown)


Julian:
There are by the way some tests that are marked pending because of a race condition in how we talk to the Lean 3 LSP on startup that I haven't been able to figure out... Specifically [these](https://github.com/Julian/lean.nvim/blob/main/lua/tests/infoview/widgets_spec.lua#L75-L137) if you didn't already see them. Pointing it out just to make sure you're aware that the Lean 3 widget code paths are less tested than they should be -- I've just locally re-enabled them and ran them just to sanity check, and they essentially pass the same way they used to after these changes (in the sense that they fail only in the same flaky way, modulo you fixing the extra linebreak), which is obviously a decent sign.

Julian:
This actually did break <Esc> in Lean 3 again.
I don't see why yet, but will have another look.

gebner:
> I don't see why yet, but will have another look.

Sorry, that's the part that I always forget to test. :-/  I've pushed a fix: 55c7ae4dce5f9f8fc883c59b9813851ca4d8f4ef

Julian:
Thanks, all good, at some point if we can get the test working it would have caught it too so not too worrying.

PR #232, opened 2022-02-11
Include the specific TComment config.

Julian:
Makes commenting inline the default, but supports block
comments via TCommentBlock.

We could also upstream this obviously.


gebner:
Thanks for doing this properly!

PR #241, opened 2022-03-11
Add basic highlighting for Lean 3 tactics.

Julian:
~~(Doesn't even bother doing this only in tactic mode -- so `have` and `let` and friends I left off the list and therefore will look a bit odd perhaps, but we can obviously improve).~~

Just copies the list from the mathlib docs.

To modify the color used (e.g. if you don't like it matching the names
of declarations), do e.g. `highlight link leanTactic Green` in your configuration.
Some other languages these days seem to be choosing to do so directly in their syntax
files rather than link to the semantic ones...

~~I didn't do Lean 4 since I was lazy and didn't read https://github.com/leanprover-community/mathlib4/blob/master/Mathlib/Mathport/Syntax.lean carefully to see if the naming is staying similar or all tactics are being ported. Obviously can do so if needed.~~

### DIFF
--- a/ftplugin/lean/lean.lua
+++ b/ftplugin/lean/lean.lua
@@ -3,7 +3,7 @@ vim.b.did_ftplugin = 1
 
 vim.opt.wildignore:append[[*.olean]]
 
-vim.opt_local.iskeyword = [[@,48-57,_,-,!,#,$,%]]
+vim.opt_local.iskeyword = [[@,48-57,_,-,.,!,#,$,%]]
 vim.opt_local.comments = [[s0:/-,mb:\ ,ex:-/,:--]]
 vim.opt_local.commentstring=[[/- %s -/]]
 
--- a/syntax/lean3.vim
+++ b/syntax/lean3.vim
@@ -71,7 +71,7 @@ syn region leanTacticMode matchgroup=Label start='\<begin\>' end='\<end\>'
     \ contains=ALLBUT,leanDeclarationName,leanEncl,leanAttributeArgs
 
 syn keyword leanKeyword end
-syn keyword leanKeyword forall fun Pi from have show assume suffices let if else then in with calc match do this
+syn keyword leanKeyword forall fun Pi from have show assume suffices let if else then in calc match do this
 syn keyword leanSort Sort Prop Type
 syn keyword leanCommand set_option run_cmd
 syn match leanCommand "#eval"
--- a/syntax/lean3.vim
+++ b/syntax/lean3.vim
@@ -9,7 +9,7 @@ syn case match
 
 syn keyword leanCommand prelude import include omit export open open_locale mutual
 syn keyword leanCommandPrefix local localized private protected noncomputable meta
-syn keyword leanModifier renaming hiding where extends using with at rec deriving
+syn keyword leanModifier renaming hiding where extends using with at only rec deriving
 
 syn keyword leanCommand namespace section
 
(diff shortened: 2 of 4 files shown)


gebner (commented):
> Doesn't even bother doing this only in tactic mode -- so have and let and friends I left off the list and therefore will look a bit odd perhaps, but we can obviously improve

This is a big problem since many tactics have short names that also occur as very common definition names.  Browsing through some files in mathlib, I see lots of false positives.

gebner:
> I didn't do Lean 4

For Lean 4 we can just wait for https://github.com/neovim/neovim/pull/15723, then we'll get accurate and complete tactic highlighting for free.

Julian:
OK, will have a look at changing `begin / end` to a region and then I think the false positives will go down. Thanks, good feedback.

syntax/lean3.vim
```diff
@@ -88,6 +119,7 @@ hi def link leanBlockComment      leanComment
 
 hi def link leanKeyword           Keyword
 hi def link leanCommand           leanKeyword
+hi def link leanTactic            Function
```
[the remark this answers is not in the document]
Julian: I think we should consider changing all these to bind to tree-sitter hlgroups, and only fallback to these when tree-sitter isn't installed. For now though followed your suggestion.

syntax/lean3.vim
```diff
@@ -24,7 +24,41 @@ syn keyword leanCommand universe universes example axioms constants
...
+        \ solve_by_elim sorry admit specialize split split_ifs squeeze_simp
+        \ squeeze_simpa squeeze_dsimp squeeze_scope subst subst_vars substs
+        \ subtype_instance success_if_fail suffices suggest swap swap_var
+        \ symmetry tautology tfae tidy trace trace_simp_set trace_state
+        \ transitivity transport triv trivial trunc_cases try type_check
+        \ unfold unfold1 unfold_cases unfold_coes unfold_projs unify_equations
+        \ use with_cases wlog zify
+        \ contained
+syn region leanTacticMode matchgroup=Label start='\<begin\>' end='\<end\>' contains=leanTactic
+syn match  leanTacticBy 'by' nextgroup=leanTactic
```
Julian, line 59: This doesn't cover braces-`by` blocks but I couldn't think how to do so easily quite yet.

Julian: Actually I guess an additional `syn match` with a region may do it now that I've thought again about it.

gebner:
This seems to work surprisingly well now.  Is it possible to disable the highlighting if a dot comes directly before the tactic name?  For example in `apply fun_like.ext` or `apply iff.intro`?

Julian:
Done.

ftplugin/lean/lean.lua
```diff
@@ -3,7 +3,7 @@ vim.b.did_ftplugin = 1
 
 vim.opt.wildignore:append[[*.olean]]
 
-vim.opt_local.iskeyword = [[@,48-57,_,-,!,#,$,%]]
+vim.opt_local.iskeyword = [[@,48-57,_,-,.,!,#,$,%]]
```
gebner, line 6: This would have been my next question, thanks for fixing this!

gebner:
Looks good to me!

Julian:
OK, let's merge and see if others have feedback then. Thanks.

ftplugin/lean/lean.lua
Julian: This is already annoying me, so I'm going to PR using `syn iskeyword`.

PR #244, opened 2022-03-17
Consider . a keyword char only for syntax highlighting.

Julian:
@gebner FYI -- you sounded like you were maybe saying you thought *with* `.` was better somehow, let me know if so? It was just me being lazy before and not figuring out how to use `syn iskeyword`. But looks like that's trivial.

(Also I removed `#`, `$` and `%` which Lean doesn't allow.)

### DIFF
--- a/ftplugin/lean/lean.lua
+++ b/ftplugin/lean/lean.lua
@@ -3,7 +3,7 @@ vim.b.did_ftplugin = 1
 
 vim.opt.wildignore:append[[*.olean]]
 
-vim.opt_local.iskeyword = [[@,48-57,_,-,.,!,#,$,%]]
+vim.opt_local.iskeyword = [[a-z,A-Z,_,48-57,192-255,!]]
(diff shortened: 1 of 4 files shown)


Julian:
> Okay, this was probably a misunderstanding and I was too cryptic. What I was happy about is that cfg.theorem etc. are now highlighted correctly in Lean 4 as well.

Ah cool OK thanks.

> There's #eval and binrel%, but # and % are also used as operators. So either classification sounds good.

Yeah, I slightly prefer `ciw` on `#eval` to give `#` ready to type `#check` say, without needing to retype the `#`. The latter I wasn't aware of, seems like `%` should maybe stay then since for operator use there'll be a space in between?

gebner:
> seems like % should maybe stay then since for operator use there'll be a space in between?

I think it should have the same status as `+`.  The `%` in macro names is really only for internal macros, so hopefully it doesn't appear too often.

Julian:
Got it, ok, back out again it goes then. Thanks.

PR #248, opened 2022-04-08
Fix state diffs.

rish987:
I'm pretty sure this was just a typo, but I'm not 100% up to speed yet so just making sure.

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -254,11 +254,16 @@ function Infoview:__was_closed()
 end
 
 --- Retrieve the contents of the infoview as a table.
-function Infoview:get_lines(start_line, end_line)
+---@param start_line number
+---@param end_line number
+---@param diff boolean @whether to get lines from the diff renderer instead
+function Infoview:get_lines(start_line, end_line, diff)
   if not self.window then error("infoview is not open") end
   start_line = start_line or 0
   end_line = end_line or -1
-  return vim.api.nvim_buf_get_lines(self.info.__renderer.buf, start_line, end_line, true)
+  local buf = diff and self.info.__diff_renderer and self.info.__diff_renderer.buf
+    or self.info.__renderer.buf
+  return vim.api.nvim_buf_get_lines(buf, start_line, end_line, true)
 end
 
 --- Toggle this infoview being open.
@@ -362,7 +367,7 @@ function Info:__set_diff_pin(params)
       use_widgets = options.use_widgets,
       parent = self
     }
-    self.__diff_renderer.__element = self.__diff_pin.__element
+    self.__diff_renderer.element = self.__diff_pin.__element
     self.__diff_pin:__show_extmark(nil, 'leanDiffPinned')
   end
 
@@ -389,7 +394,7 @@ function Info:__clear_diff_pin()
   if not self.__diff_pin then return end
   self.__diff_pin:__teardown()
   self.__diff_pin = nil
-  self.__diff_renderer.__element = self.pin.__element
+  self.__diff_renderer.element = self.pin.__element
   self:render()
 end
 
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -257,14 +257,20 @@ end
 --- Retrieve the contents of the infoview as a table.
 ---@param start_line number
 ---@param end_line number
----@param diff boolean @whether to get lines from the diff renderer instead
-function Infoview:get_lines(start_line, end_line, diff)
+function Infoview:get_lines(start_line, end_line)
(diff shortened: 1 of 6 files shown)


Julian:
Oy, are they totally broken at the minute? Any chance you care to attempt to add a test? Otherwise I can have a look.

lua/tests/infoview/pin_spec.lua
```diff
@@ -278,34 +283,44 @@ describe('infoview pins', helpers.clean_buffer('lean', dedent[[
         h37 : p
         ⊢ q ∨ p
       ]], vim.api.nvim_buf_get_name(0)))
```
rish987, line 285: (this was a duplicate assertion I believe)

lua/lean/infoview.lua
```diff
@@ -254,11 +254,16 @@ function Infoview:__was_closed()
 end
 
 --- Retrieve the contents of the infoview as a table.
-function Infoview:get_lines(start_line, end_line)
+---@param start_line number
+---@param end_line number
+---@param diff boolean @whether to get lines from the diff renderer instead
```
Julian, line 259: I'm not a fan of passing bools to functions. Can you make a separate `get_diff_lines` function, with a helper that both functions call if it indeed is needed?

lua/lean/infoview.lua
```diff
@@ -362,7 +367,7 @@ function Info:__set_diff_pin(params)
       use_widgets = options.use_widgets,
       parent = self
     }
-    self.__diff_renderer.__element = self.__diff_pin.__element
+    self.__diff_renderer.element = self.__diff_pin.__element
```
Julian, line 370: Can we instead keep this private and fix the place looking for it?

rish987: Do you mean to make `element` a private member of `BufRenderer`? (not sure I'm understanding correctly)

lua/lean/infoview.lua
rish987: Sounds good, what do you think about moving this to `helpers.lua` as well? Just did a search and looks like we're only using this for testing anyways.

Julian: So I think for now I've left it here because it's also useful interactively (and later programmatically for anything outside of this that wants to see what's in the infoview until we need a more structured way of passing data out of it), so it just seemed like decent API to expose even if it wasn't used internally. `get_diff_lines` seems less likely to be useful to end-users I suppose but if only for symmetry it seems fine to put here too to me.

lua/tests/helpers.lua
```diff
@@ -152,15 +152,18 @@ end
 assert:register('assertion', 'contents', has_buf_contents)
 
 --- Assert about the current infoview contents.
-local function has_infoview_contents(_, arguments)
-  local expected = dedent(arguments[1][1] or arguments[1])
-  local target_infoview = arguments[1].infoview or infoview.get_current_infoview()
-  local got = table.concat(target_infoview:get_lines(), '\n')
-  assert.are.same(expected, got)
-  return true
+local function has_infoview_contents(diff)
```
Julian, line 155: (Same here on boolean arguments, I'd prefer this was 2 separate functions which possibly call a helper function that does most of the work)

lua/tests/infoview/pin_spec.lua
Julian: In the original code it's meant to be a "guard assertion" -- meaning it was here so that if someone introduces a test before or after this one and doesn't realize tests are "dirty" and leak state from one to the other, this one should now fail in a more transparent way (by saying "the initial contents of the infoview are different than expected"). But let me know if you still think it looks duplicated, I didn't fully think about how useful the guard is here.

lua/lean/infoview.lua
Julian: So, I don't like doing `foo.bar = something` from outside of `foo`, I prefer immutable (or really persistent) objects which don't have their attributes changed over their lifetimes. We don't have that now, so we do need to do this in places, but if we do do this I'd like it to be to private attributes so it's even more clear it's something intended for change. Even better would be if we could create `__diff_renderer` here in the first place rather than having an existing one and mutating one of its attributes. Not sure if that's feasible or not or if there's a reason it needs to exist before this point in `Info`'s lifecycle.

I can look closer though if the above is more confusing than helpful (which it may be...).

rish987: Just made it so that we use an element wrapper on top of the diff pin element, so we only need to modify its children (via a public function) similarly to how we currently do for the pins element. Let me know if it looks any better now.

Julian:
Merging after some minor tweaks, thanks for the fix (and sorry it broke).

PR #250, opened 2022-04-15
Add helper for better test synchronization.

rish987:
ISTM that, rather than synchronizing with `wait_for_infoview_contents(...)` to wait for the contents to match after a pin update, we can instead just wait until the infoview's pins are no longer loading or processing. I've only ported `pin_spec.lua` for now as a proof-of-concept, just wanted to first make sure we agree that this is a better way of doing this before going ahead with the full port.

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -707,7 +707,6 @@ function Pin:__started_loading()
   if self.loading then return false end
   self.loading = true
   self.__element:set_children{ self.__data_element }
-  self:__render_parents()
   return true
 end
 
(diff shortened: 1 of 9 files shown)


Julian:
Nice. This definitely looks nicer -- though can it be even nicer if it's just done in `assert.infoview_contents` automatically?

Julian:
@rish987 this looks really nice I think, and more reliable! Going to merge immediately I think. I saw you said you had more in mind, obv send a follow-up with anything else? And well done as usual this is great.

rish987:
Yeah it's still going to be a bit flaky since the fileProgress notifications seemed unreliable in Lean 3 (sometimes the infoview would get data before we got a single notification). Will follow up on it.

Julian:
Ah, fun, ok, well it still seems in my unscientific tests to pass more often than before, so progress...

Issue #252, opened 2022-04-19
Make it easier to notice when a document is still processing for diagnostics


Julian:
Right now, running e.g. `vim.diagnostic.goto_next()` will silently succeed if the file is still processing (and therefore has no diagnostics).

I'm not sure we can fully hook into doing this, but it might be nice if there was an easy way to have things fail until/unless the document is fully processed, as I (and probably others) sometimes use `vim.diagnostic.goto_next` as a poor-man's "is there anything to fix in this file".

Julian:
Ok erroring I think is indeed not possible without a wrapper but the obvious solution here is probably to just set a diagnostic at the top of the processing region (and move/remove it as processing continues).

PR #253, opened 2022-04-28
Show file progress as a diagnostic while it occurs.

Julian:
Helps make sure someone using e.g. vim.diagnostic.goto_next
notices that the file isn't ready yet.

Closes: https://github.com/Julian/lean.nvim/issues/252


Julian:
(The tests seem a bit flaky again locally, will investigate, but pinging the both of you in case you have an opinion on how well this works for you.)

lua/lean/_util.lua
```diff
@@ -12,6 +12,20 @@ function M.tbl_repeat(value, times)
...
 
+--- Fetch the diagnostics for all Lean LSP clients from the current buffer.
+function M.lean_lsp_diagnostics(opts, bufnr)
+  bufnr = bufnr or 0
+  opts = opts or {}
+  local diagnostics = {}
+  for _, client in pairs(vim.lsp.buf_get_clients(bufnr or 0)) do
+    if client.name:match('^lean') then
+      opts.namespace = vim.lsp.diagnostic.get_namespace(client.id)
+      vim.list_extend(diagnostics, vim.diagnostic.get(0, opts))
```
rish987, line 23:
```suggestion
      vim.list_extend(diagnostics, vim.diagnostic.get(bufnr, opts))
```

Julian: Nice catch, I'd just noticed this myself, fixing, but first fighting with trying to get `vim.tbl_map` to do what I want.

Julian: (Spoiler: I gave up.)

lua/lean/_util.lua
```diff
@@ -12,6 +12,19 @@ function M.tbl_repeat(value, times)
...
 end
 
+--- Fetch the diagnostics for all Lean LSP clients from the current buffer.
+function M.lean_lsp_diagnostics(opts, bufnr)
+  opts = opts or {}
+  local diagnostics = {}
+  for _, client in pairs(vim.lsp.buf_get_clients(bufnr)) do
+    if client.name:match('^lean') then
+      opts.namespace = vim.lsp.diagnostic.get_namespace(client.id)
+      vim.list_extend(diagnostics, vim.diagnostic.get(bufnr, opts))
```
rish987, line 22:
```suggestion
      vim.list_extend(diagnostics, vim.diagnostic.get(bufnr or 0, opts))
```
I think you need an explicit first parameter for the second to be considered.

lua/lean/_util.lua
rish987: Huh, what exactly was the issue?

lua/lean/_util.lua
Julian: Thanks!

rish987:
LGTM as well, thank you! Yeah the tests in general need some work (I've been busy reading lately but will get back to #250 soon).

lua/lean/_util.lua
Julian: I couldn't get the tests to pass after changing to:

```lua
function M.lean_lsp_diagnostics(opts, bufnr)
  bufnr = bufnr or 0
  opts = opts or {}
  return vim.tbl_flatten(vim.tbl_map(function(client)
    if not client.name:match('^lean') then return {} end
    opts.namespace = vim.lsp.diagnostic.get_namespace(client.id)
    return vim.diagnostic.get(bufnr, opts)
  end, vim.lsp.buf_get_clients(bufnr)))
end
```

and got fed up fighting with Lua.

PR #258, opened 2022-05-19
Go-to-definition in Lean 4 widgets

gebner:
Depends on https://github.com/leanprover/lean4/pull/1157


lua/lean/infoview.lua
```diff
@@ -35,6 +35,9 @@ local options = {
     mappings = {
       ['K'] = 'click',
       ['<CR>'] = 'click',
+      ['gd'] = 'go_to_def',
+      ['gD'] = 'go_to_decl',
+      ['gt'] = 'go_to_type',
```
Julian, line 40: `gt` is a somewhat useful existing mapping ([for `:tabnext`](https://neovim.io/doc/user/tabpage.html#gt)). `gy` is unused and maybe a good alternative if we don't want to relegate this to `<localleader>`.

Julian (commented):
Nice!

gebner:
> heading to toZero : Zero M₀, on Zero, gd and gt do nothing

You should be able to `gd` on `Zero.zero`.  Any function application, like `Zero M₀`, [doesn't work yet](https://github.com/leanprover/lean4/pull/1157#issuecomment-1131498587).

Julian:
Ah sorry, I misread that comment previously. Got it, thanks.

PR #259, opened 2022-05-20
Add note for vim non-experts about LocalLeader.

langfield:
Add a note about how to remap `<LocalLeader>` before the key mappings table in the README.

I am by no means an expert with vim/neovim. Found the use of `<LocalLeader>` in the mappings table very confusing on first look, and the document that the `:help <LocalLeader>` command brings up is massive (over 1500 lines) and mostly irrelevant to using `lean.nvim` out of the box.

Feedback appreciated, not completely sure if this is useful.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -149,7 +149,7 @@ either within Lean source files or within Infoview windows:
 In Lean Files
 ^^^^^^^^^^^^^
 
-The key binding ``<LocalLeader>`` below refers to a configurable prefix key 
+The key binding ``<LocalLeader>`` below refers to a configurable prefix key
 within vim (and neovim). You can check what this key is set to within neovim
 by running the command ``:echo maplocalleader``. An error like
 ``E121: Undefined variable: maplocalleader`` indicates that it may not be set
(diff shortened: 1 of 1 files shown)


Julian:
Thanks! I'm super happy to add this sort of thing, definitely don't want to assume the audience isn't a vim expert!

That said, there's this just below the table:

> Note

> See :help <LocalLeader> if you haven't previously interacted with the local leader key. Some vim users remap this key to make it easier to reach, so you may want to consider what key that means for your own keyboard layout. My (Julian's) <Leader> is set to <Space>, and my <LocalLeader> to <Space><Space>, which may be a good choice for you if you have no other preference.

Did you see that note and still find it unclear? Or not see it and think it's better before the table rather than after? Or other?

Regardless thanks for any suggestion on making this clearer!

langfield:
Thanks for replying so quickly! Yes I did see that note, and was still a bit confused. Nothing some googling wouldn't remedy, of course, but I think the more self-contained the installation is, the better.

Julian:
Got it, sounds good -- appreciate the PR!

Issue #260, opened 2022-05-25
Lean 4 server crashing


rish987:
Hey, I've just started using Lean 4 and often get the error:

`Client 1 quit with exit code 1 and signal 0`

when I start editing. Not sure if it's related to #203? Just wanted to ask if you've ever seen this and know what's up, if not LMK if there's any more info I can provide to help find the issue. I didn't see any output in `~/.cache/nvim/{log,lsp.log}`, though I could also try to get a core dump.

Julian:
I don't think I've seen this before but I've used Lean4 less recently -- what version are you on just to be sure? And nothing on stderr when this happens?

rish987:
`leanprover/lean4:nightly-2022-05-18` (sorry, that should have been the first thing I included here)

Yeah I don't see any other output, the server just silently fails. Since you haven't seen this it very well may be my problem, I'll try things within an isolated environment tomorrow and get back to you.

gebner:
This looks like your Lean installation is messed up a bit.  When you start lean from nvim, it runs `lean --server`, which in turns runs one `lean --worker` process per file.  The error means that these two processes come from different Lean versions.

rish987:
Aha, it's completion related. I'm using `nvim-cmp` and see a crash (with the watchdog error above) inconsistently when invoking completion and then entering a newline. Not sure where the error lies, but I guess I can do without completion for now until I have time to look deeper.

Julian:
I've now actually seen this (on `Lean (version 4.0.0, commit 3eeeca22e281, Release)`) where I got a notification that:

```
2022-06-27T11:57:27  WARN Client 1 quit with exit code 1 and signal 0
```

but nothing on stderr and nothing else in `:messages`. I was doing nothing in particular special while that happened (just editing some unremarkable structure).

It should be that this is an upstream issue though right, so we should probably take it there if/when we diagnose this.

rish987:
OK good to know, yes I can confirm that I'm still seeing crashes as well without autocompletion, but they've become much less frequent. Hopefully the problem with completion is related because it's pretty easy for me to reproduce.

rish987:
So I think this could be coupled with synchronization issues (i.e. the server not picking up changes correctly) which I've noticed as well. I suspect the issue is on the nvim side, will investigate.

rish987:
[...]

PR #264, opened 2022-07-01
Update infoview after opening.

gebner:
When you manually open the infoview, it used to start up blank.  This triggers an update right after it is opened.

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -1112,6 +1112,7 @@ function infoview.toggle()
   else
     infoview.open()
   end
+  infoview.__update()
 end
 
 --- Toggle whether the current pin receives updates.


Julian:
Added a test, and made this happen on all open, not just toggle, but definitely seems reasonable to me. (Let me know if you disagree with the 2 commits obviously, but seems unlikely).

The CI failure on nightly is annoying, I actually think we may be triggering #260 in CI (and this new test means I can sometimes trigger it locally now, but not deterministically). Will merge though if only nightly fails as that's nothing new to this PR :/

lua/tests/infoview/autoopen_false_spec.lua
```diff
@@ -30,5 +30,18 @@ describe('infoview', function()
...
     infoview.open()
     assert.windows.are(lean_window, infoview.get_current_infoview().window)
+    infoview.close()
+  end)
+
+  it('shows infoview contents immediately after they are opened', function(_)
+    vim.cmd('edit! ' .. fixtures.lean_project.path .. '/Test.lean')
+    assert.windows.are(lean_window)
+    infoview.open()
+    helpers.move_cursor{ to = {3, 27} }
```
gebner, line 40:
```suggestion
    helpers.move_cursor{ to = {3, 27} }
    infoview.open()
```

The other order used to work already.

Julian: Thanks, good catch! Though that worried me to double check, and the test passes even without the change still, so I made some other error in it. Will look in a sec.

Julian: OK now it fails without and passes with. I honestly don't know why what I did now is different than before :/ but there's some digging that needs doing later.

Issue #262, opened 2022-06-27
Add an equivalent of refresh file dependencies for Lean 4


Julian:
We need a command equivalent of [lean4.refreshFileDependencies](https://github.com/leanprover/vscode-lean4/blob/dc33aae89c8b89e5febc9337a8e1136126ee5dec/vscode-lean4/src/leanclient.ts#L434-L462) for telling the Lean server to restart the worker for the current file / refresh building imported files.

gebner:
I've been abusing `:e` for this.

Julian:
Probably we should keep telling others that's the "official" way to do it -- the reason to have something else just in case would be for someone (yesterday me) having a `BufRead` autocmd they're running, in which case hitting `:e` will keep triggering it unnecessarily.

Maybe doesn't have to be an exact equivalent of refresh file dependencies, could just be we kill + restart an LSP client.

gebner:
Yeah, the `:e` is not ideal.  You also need to do `:w` first.

The server expects a didClose followed by a didOpen notification to restart the file worker.  Restarting the LSP server is useful as well (and unfortunately not as easy as `:LspRestart`), but it restarts all the file workers, which is rarely what you want.

PR #265, opened 2022-07-04
Add an equivalent of lean4.refreshFileDependencies.

Julian:
Needed for forcibly telling Lean 4 to rebuild dependencies when
editing a dependent file.

:e works but requires writing the file, and will refire Buf-related
events, which may be undesirable.

Ultimately Lean 4 is hopefully planning to obviate the need to do this
manually, but for now this mirrors VSCode.

Closes: #262

### DIFF
--- a/lua/lean/lsp.lua
+++ b/lua/lean/lsp.lua
@@ -85,4 +85,21 @@ function lsp.handlers.diagnostics_handler (_, params)
   require"lean.infoview".__update_pin_by_uri(params.uri)
 end
 
+---Refresh file dependencies for an open Lean 4 file.
+---See e.g. https://github.com/leanprover/lean4/blob/master/src/Lean/Server/README.md#recompilation-of-opened-files
+---@param bufnr? number
+function lsp.refresh_file_dependencies(bufnr)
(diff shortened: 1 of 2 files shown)


lua/lean/lsp.lua
```diff
@@ -85,4 +85,21 @@ function lsp.handlers.diagnostics_handler (_, params)
...
+  bufnr = bufnr or 0
+  local client = lsp.get_lean4_server(bufnr)
+
+  local uri = vim.uri_from_bufnr(bufnr)
+  client.notify('textDocument/didClose', { textDocument = { uri = uri } })
+  client.notify('textDocument/didOpen', { textDocument = {
+      version = 0,
+      uri = uri,
+      languageId = client.config.get_language_id(bufnr, vim.bo.filetype),
+      text = table.concat(vim.api.nvim_buf_get_lines(bufnr, 0, -1, true), '\n')
```
gebner, line 101: This will break with non-unix line-endings.  Maybe we should just copy the `buf_get_full_text` function from neovim.

Julian: Bleh, seems so, thanks -- done.

PR #268, opened 2022-08-09
WIP: `hop.nvim` integration

rish987:
Depends on (minor) bug fixes in phaazon/hop.nvim#304. What I've done so far is to map:

- `<localleader>jj` to hop to interactive elements of the infoview
- `<localleader>jd` to use the hop interface to select an element of the infoview that supports go-to-definition (_restricting the hints to such elements_), and go to that definition.
- `<localleader>jh` to use the hop interface to select an interactive element to click (though this is broken at the moment in that the windows don't show up, it's a UI bug I believe)

All of the above operate _from the source code_, so you don't have to context-switch to the infoview window and then go back to the original window. There are a number of additional things I want to do (will make a TODO list later) but I wanted to put this PR up now in case there's any early feedback.

[asciicast](https://asciinema.org/a/hXoWQQGCYRmDimiD9q3pR7ZMW)

^ Sorry if that's a bit inscrutable, I'm trying to figure out how to get my terminal to show keystrokes, but essentially what I did was `<localleader>jj` to go to `some` in the infoview and see its type, and then `<localleader>jd` to go-to-definition of `Option`.

Side note: If you're wondering why I haven't been super active lately, it's because I've been ramping up at a new job where I get to work on Lean full-time! I've settled pretty well at this point and am now finding a healthy balance between work, contributing to neovim, and learning about the Lean 4 implementation with the goal of eventually contributing to its usability aspects. Thanks for your patience!!!

### DIFF
--- a/lua/lean/infoview/components.lua
+++ b/lua/lean/infoview/components.lua
@@ -144,6 +144,15 @@ local function code_with_infos(t, sess)
       end
     end
 
+    --FIXME this is essentially a thunked boolean field, adding it as an event
+    --is not exactly, but I'm not sure where to stick it at the moment
+    ---@param kind GoToKind
+    local has_go_to = function(_, kind)
+      local links, err = sess:getGoToLocation(kind, info_with_ctx)
+      if err or #links == 0 then return false end
+      return true
+    end
+
     ---@param kind GoToKind
     local go_to = function(_, kind)
       local links, err = sess:getGoToLocation(kind, info_with_ctx)
@@ -173,6 +182,7 @@ local function code_with_infos(t, sess)
       clear = function(ctx) if info_open then do_reset(ctx) end end,
       go_to = go_to,
       go_to_def = go_to_def,
+      has_go_to = has_go_to,
       go_to_decl = go_to_decl,
       go_to_type = go_to_type,
     }
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -1188,6 +1188,42 @@ function infoview.go_to()
   end
 end
 
+--- Hop to an interactive infoview element by moving to that element.
+function infoview.hop()
+  local curr_iv = infoview.open()
+  local curr_info = curr_iv.info
+  local opts = {}
+  opts.jump_on_sole_occurrence = false
+  if vim.api.nvim_get_current_win ~= curr_iv.window then
+    opts.multi_windows = true
+  end
+  curr_info.__renderer:hop_to(opts)
+end
+
+--- Hop to an interactive infoview element by going to its definition.
+function infoview.hop_definition()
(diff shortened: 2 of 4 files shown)


rish987:
Also, obviously this relates to #197, we're lucky that @phaazon has since made an official API for us to use! Again, the fixes in #304 should stay minor.

lua/lean/infoview/components.lua
```diff
@@ -144,6 +144,15 @@ local function code_with_infos(t, sess)
       end
     end
 
+    --FIXME this is essentially a thunked boolean field, adding it as an event
+    --is not exactly, but I'm not sure where to stick it at the moment
+    ---@param kind GoToKind
+    local has_go_to = function(_, kind)
+      local links, err = sess:getGoToLocation(kind, info_with_ctx)
```
gebner, line 151: This will be very slow if you call it sequentially on all subterms in the infoview.  I'd just add a hint everywhere (no matter whether `getGoToLocation` succeeds or not).

Julian:
Cool! Screenshare looks nice, no concerns from me!

Only minor thing is perhaps just put binding those keys behind an `if` which only runs when `hop.nvim` is installed, so we don't quash a keybind someone may be using for something else entirely without hop.

lua/lean/infoview/components.lua
rish987: Yes, this is a temporary hack until we add some kind of lightweight indicator of which elements have go-to defintition/declaration/etc. on the server side. hop.nvim orders hints by reachability on your keyboard, so from an ergonomic perspective I don't think I can settle for less lol. Even this hacked solution has been very fast for me though? If you're noticing performance issues we could of course make it configurable, but even so personally I'd rather wait a bit longer for the hints than have to press an additional/harder-to-reach key.

rish987:
TODO:

- [ ] fix `:LeanHopInfoviewHover`
- [ ] allow for the user to provide lean-specific [hop configuration options](https://github.com/phaazon/hop.nvim/wiki/Configuration) in the `lean.nvim` configuration table that override the options they have globally set for `hop.nvim`
- [ ] figure out how to deal with overlapping hints that result from notation, e.g. go-to-definition for `Nat × Nat` has a hint for each `Nat`, however the `· × ·` hint is lost entirely as its hint range starts at the same column as the first `Nat`
- [ ] properly check that `hop.nvim` is installed before setting up any related functionality (keymaps, commands, etc.)
- [ ] make the hop commands work in the infoview itself with similar keymaps
- [ ] on the `hop.nvim` side, add an option to support restricted dimming to only windows where hints are present (right now the `multi_windows` option causes all windows to be dimmed)

Obviously some of these will be more involved than others and so are not all for this PR, just wanted to make a list somewhere that I can keep track of.

PR #269, opened 2022-08-14
Add an option to open the infoview in an entirely new tab.

tage64:
This pull request adds an option (infoview.separate_tab: bool) which if set to true will cause the infoview to be opened in a new tabpage rather then in a horizontal or vertical split of the current window.

This is probably not useful for most users since one will not be able to see real time changes in the infoview. But I'm using a [screen reader](https://en.wikipedia.org/wiki/Screen_reader) which speaks all dynamic changes in the terminal immediately. And it sometimes become too much information at the same time when it speaks both changes in the main buffer and the infoview.

A better solution would of course be to make the screen reader smarter so that it can recognize horizontally/verticly split windows and only speak changes in the current window. But that was too complicated for a Sunday evening so I decided to change this pluggin instead.

I'm not an expert in Lua, so feel free to comment or improve my changes.

I would understand if you think the audience for this feature is very limited. It might be that there will only be me and other visually impared people who make use of this feature. But the changes are very small though, so I hope you have no problems to merge it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -313,6 +313,12 @@ Full Configuration & Settings Information
         -- top | bottom
         horizontal_position = "bottom",
 
+        -- Always open the infoview window in a separate tabpage.
+        -- Might be useful if you are using a screen reader and don't want too
+        -- many dynamic updates in the terminal at the same time.
+        -- Note that `height` and `width` will be ignored in this case.
+        separate_tab = false,
+
         -- Show indicators for pin locations when entering an infoview window?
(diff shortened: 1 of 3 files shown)


rish987 (approved):
Nice, thanks! Added some tests. I could also see myself finding this useful for distraction-free Lean. Judging from this PR, it shouldn't be too difficult for us to extend the infoview viewing options even further, feel free to make an issue/another PR if you have any requests!

PR #273, opened 2022-09-02
Perform ftdetect in pure lua with no plugin call

### LINKED ISSUE #272: ftdetect script references rest of plugin
Hi!
I'd like to only load the `lean.nvim` plugin if im editing a lean file. I'm currently using the packer package manager, so I did
```lua
 use { 'Julian/lean.nvim', ft = { 'lean', 'lean3' }}
```

In order to do this, packer installs `lean.nvim`'s ftdetect script _without_ the rest of the plugin, so that `lean.nvim` can do filetype detection in order to determine if the plugin should be loaded. Unfortunately, `lean.nvim`'s [ftdetect script](https://github.com/Julian/lean.nvim/blob/main/ftdetect/lean3.vim) calls into the plugin itself. Because the plugin isn't loaded at the time, this causes filetype detection to crash.

Getting around this isn't too hard, since the fix is to just do
```lua
 use { 'Julian/lean.nvim', ft = { 'lean', 'lean3' }, module = 'lean'}
```
but it still doesn't seem optimal since this is loading the entire lean plugin when it's not necessary to do so.

It seems like it would be pretty easy to solve this by moving [`main/lua/lean/ft.lua`](https://github.com/Julian/lean.nvim/blob/main/lua/lean/ft.lua) to `ftdetect/lean3.lua` instead of using the vimscript shim.

4e554c4c:
Summary:
I use the packer package manager which lets one only enable certain
plugins if a filetype related to them is set. In order to get around the
issue that these plugins also may also _set_ those filetypes---packer
installs and loads only the filedetect scripts of the `ft` triggered
plugins. If one attempts to do this for lean.nvim, then the filedetect
script fails as it immediately calls into the lean.nvim plugin which is
yet to be loaded.
Since filedetect scripts can be pure lua in neovim, i have replaced the
`ftdetect/lean3.vim` shim with the standalone `ftdetect/lean.lua`
script.
In order to get around configuration issues, this filedetect script
configures the `ft` configuration global which can be modified by the
plugin itself.

Test Plan:
edited `lua/tests/fixtures/example-lean4-project/Test.lean` and
`lua/tests/fixtures/example-lean3-project/src/foo.lean` and verified
that `set ft?` returned `lean` and `lean3` respectively.

before:
```
$ TEST_FILE=lua/tests/ft_spec.lua ./lua/tests/scripts/run_tests.sh

========================================
Testing: 	lua/tests/ft_spec.lua
Success	||	ft.detect detects nonexisting lean 4 files
Success	||	ft.detect detects nested nonexisting lean 4 files
Success	||	ft.detect detects existing lean 4 files
Success	||	ft.detect detects nested existing lean 4 files
Success	||	ft.detect detects nonexisting lean 3 files
Success	||	ft.detect detects nested nonexisting lean 3 files
Success	||	ft.detect detects existing lean 3 files
Success	||	ft.detect detects nested existing lean 3 files
Fail	||	ft.detect detects lean 4 standard library files
            lua/tests/ft_spec.lua:30: attempt to call field 'wait_for_loading_pins' (a nil value)

            stack traceback:
            	lua/tests/ft_spec.lua:30: in function <lua/tests/ft_spec.lua:24>

Fail	||	ft.detect marks lean 4 standard library files nomodifiable by default
            lua/tests/ft_spec.lua:42: Didn't jump to core Lean!
            Expected to be truthy, but value was:
            (nil)

            stack traceback:
            	lua/tests/ft_spec.lua:42: in function <lua/tests/ft_spec.lua:39>

Success	||	ft.detect does not mark other lean 4 files nomodifiable
Fail	||	ft.detect detects lean 3 standard library files
            lua/tests/ft_spec.lua:57: attempt to call field 'wait_for_line_diagnostics' (a nil value)

            stack traceback:
            	lua/tests/ft_spec.lua:57: in function <lua/tests/ft_spec.lua:51>

Fail	||	ft.detect marks lean 3 standard library files nomodifiable by default
            lua/tests/ft_spec.lua:68: Expected to be truthy, but value was:
            (nil)

            stack traceback:
            	lua/tests/ft_spec.lua:68: in function <lua/tests/ft_spec.lua:66>

Success	||	ft.detect does not mark other lean 3 files nomodifiable

Success: 	10
Failed : 	4
Errors : 	0
========================================
Tests Failed. Exit: 1
```
after:
```
$ TEST_FILE=lua/tests/ft_spec.lua ./lua/tests/scripts/run_tests.sh

========================================
Testing: 	lua/tests/ft_spec.lua
Success	||	ft.detect detects nested nonexisting lean 4 files
Success	||	ft.detect detects existing lean 4 files
Success	||	ft.detect detects nested existing lean 4 files
Success	||	ft.detect detects nonexisting lean 4 files
Success	||	ft.detect detects nested nonexisting lean 3 files
Success	||	ft.detect detects existing lean 3 files
Success	||	ft.detect detects nested existing lean 3 files
Success	||	ft.detect detects nonexisting lean 3 files
Fail	||	ft.detect detects lean 4 standard library files
            lua/tests/ft_spec.lua:30: attempt to call field 'wait_for_loading_pins' (a nil value)

            stack traceback:
            	lua/tests/ft_spec.lua:30: in function <lua/tests/ft_spec.lua:24>

Fail	||	ft.detect marks lean 4 standard library files nomodifiable by default
            lua/tests/ft_spec.lua:42: Didn't jump to core Lean!
            Expected to be truthy, but value was:
            (nil)

            stack traceback:
            	lua/tests/ft_spec.lua:42: in function <lua/tests/ft_spec.lua:39>

Success	||	ft.detect does not mark other lean 4 files nomodifiable
Fail	||	ft.detect detects lean 3 standard library files
            lua/tests/ft_spec.lua:57: attempt to call field 'wait_for_line_diagnostics' (a nil value)

            stack traceback:
            	lua/tests/ft_spec.lua:57: in function <lua/tests/ft_spec.lua:51>

Fail	||	ft.detect marks lean 3 standard library files nomodifiable by default
            lua/tests/ft_spec.lua:68: Expected to be truthy, but value was:
            (nil)

            stack traceback:
            	lua/tests/ft_spec.lua:68: in function <lua/tests/ft_spec.lua:66>

Success	||	ft.detect does not mark other lean 3 files nomodifiable

Success: 	10
Failed : 	4
Errors : 	0
========================================
Tests Failed. Exit: 1
```
same number of tests pass and fail.

Fixes #272

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,5 +1,7 @@
 globals = {
     "_clean_buffer_count",
     "vim",
+    "lean_nvim_ft_options",
+    "_LEAN3_STANDARD_LIBRARY",
     assert = {fields = {"message"}},
 }
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -2,6 +2,5 @@ globals = {
     "_clean_buffer_count",
     "vim",
     "lean_nvim_ft_options",
-    "_LEAN3_STANDARD_LIBRARY",
     assert = {fields = {"message"}},
 }
--- a/ftdetect/lean.lua
+++ b/ftdetect/lean.lua
@@ -1,0 +1,65 @@
+_LEAN3_STANDARD_LIBRARY = '.*/[^/]*lean[%-]+3.+/lib/'
+local _LEAN3_VERSION_MARKER = '.*lean_version.*".*:3.*'
+local _LEAN4_VERSION_MARKER = '.*lean_version.*".*lean4:.*'
+
+local find_project_root = require('lspconfig.util').root_pattern(
+  'leanpkg.toml',
+  'lakefile.lean',
+  'lean-toolchain'
+)
+
+lean_nvim_ft_options = {
+  default = "lean",
+  nomodifiable = {
+    '.*/src/lean/.*',       -- Lean 4 standard library
+    '.*/lib/lean/src/.*',   -- Lean 4 legacy standard library
+    '.*/lean_packages/.*',  -- Lean 4 dependencies
+    _LEAN3_STANDARD_LIBRARY .. '.*',
+    '/_target/.*/.*.lean'   -- Lean 3 dependencies
+  }
+}
+
+local function detect(filename)
+  if filename:match('^fugitive://.*') then
+    filename = pcall(vim.fn.FugitiveReal, filename)
+  end
+
+  local abspath = vim.fn.fnamemodify(filename, ":p")
+  local filetype = lean_nvim_ft_options.default
+  if abspath:match(_LEAN3_STANDARD_LIBRARY) then
+    filetype = 'lean3'
+  else
+    local project_root = find_project_root(abspath)
+    local succeeded, result
+    if project_root then
+      succeeded, result = pcall(vim.fn.readfile, project_root .. '/lean-toolchain')
+      if succeeded then
+        if result[1]:match('.*:3.*') then filetype = 'lean3'
+        elseif result[1]:match('.*lean4:.*') then filetype = 'lean'
+        end
+      else
+        succeeded, result = pcall(vim.fn.readfile, project_root .. '/leanpkg.toml')
+        if succeeded then
+          for _, line in ipairs(result) do
+            if line:match(_LEAN3_VERSION_MARKER) then
+              filetype = 'lean3'
+              break
+            end
+            if line:match(_LEAN4_VERSION_MARKER) then
+              filetype = 'lean'
+              break
+            end
+          end
+        end
+      end
+    end
+  end
+  vim.opt.filetype = filetype
+end
+
+vim.api.nvim_create_autocmd({'BufRead','BufNewFile'}, {
+    pattern = '*.lean',
+    callback = function(opts)
+      detect(opts.file)
+    end,
+})
--- a/ftdetect/lean.lua
+++ b/ftdetect/lean.lua
@@ -1,4 +1,4 @@
-_LEAN3_STANDARD_LIBRARY = '.*/[^/]*lean[%-]+3.+/lib/'
+local _LEAN3_STANDARD_LIBRARY = '.*/[^/]*lean[%-]+3.+/lib/'
 local _LEAN3_VERSION_MARKER = '.*lean_version.*".*:3.*'
 local _LEAN4_VERSION_MARKER = '.*lean_version.*".*lean4:.*'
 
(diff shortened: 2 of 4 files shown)


.luacheckrc
```diff
@@ -1,5 +1,7 @@
 globals = {
     "_clean_buffer_count",
     "vim",
+    "lean_nvim_ft_options",
+    "_LEAN3_STANDARD_LIBRARY",
```
Julian, line 5: This doesn't have to be a global I think, there's just an accidental removal of the `local` in `ftdetect/lean`, if you add it back you can remove this I suspect.

ftdetect/lean.lua
```diff
@@ -0,0 +1,65 @@
...
+  }
+}
+
+local function detect(filename)
+  if filename:match('^fugitive://.*') then
+    filename = pcall(vim.fn.FugitiveReal, filename)
+  end
+
+  local abspath = vim.fn.fnamemodify(filename, ":p")
+  local filetype = lean_nvim_ft_options.default
```
Julian, line 28: I think this breaks someone changing the ft default (and probably we don't have a test covering that). In other words, we support changing what to do with single file `.lean` files from lean 4 to lean 3 via a config option, and I think that setting may be being ignored now. But that's just from reading this mentally. Can you double check?

Julian:
Thanks! Left 2 small comments, appreciate the PR!

ftdetect/lean.lua
Julian: This is the config option (from the README):

>   ft = {
    -- What filetype should be associated with standalone Lean files?
    -- Can be set to "lean3" if you prefer that default.
    -- Having a leanpkg.toml or lean-toolchain file should always mean
    -- autodetection works correctly.
    default = "lean",

4e554c4c: nope! this still works
how i understand it is that if you choose to initialize the plugin you can still change the behaviour before ft detection runs
e.g. the control flow goes something like
1. ftdetect script loaded
2. plugin loaded
3. autocmd for ftdetect runs

I confirmed this by setting `ft = { default = "lean5", }`, and then edited a `test.lean` empty file, then `set ft?` showed `lean5` as the filetype.

The main issue is that if you _don't_ load the plugin (e.g. don't call `require 'lean'.setup` during initialization) then you won't be able to configure the default at all.
We could get around this by making the `lean_nvim_ft_options` global a part of the public API, and extend it in the ftdetect script instead of overwriting it---but that's up to you.

Julian:
Perfect! Thanks for double checking. This looks great now, merging!

PR #276, opened 2022-09-05
Fix crashing when loading lean config

### LINKED ISSUE #275: attempt to index global 'lean_nvim_ft_options' (a nil value)
After some recent updates, the following error jumps out when I re-launch neovim.

```text
Error detected while processing /home/user/.config/nvim/init.lua:
E5113: Error while calling lua chunk: ...re/nvim/site/pack/packer/start/lean.nvim/lua/lean/ft.lua:3: attempt to index global 'lean_nvim_ft_options' (a nil value)
stack traceback:
...re/nvim/site/pack/packer/start/lean.nvim/lua/lean/ft.lua:3: in main chunk
[C]: in function 'require'
.../nvim/site/pack/packer/start/lean.nvim/lua/lean/init.lua:61: in function 'setup'
/home/user/.config/nvim/lua/config/lsp.lua:95: in main chunk
[C]: in function 'require'
/home/user/.config/nvim/init.lua:4: in main chunk
```

To reproduce, use this `init.lua` config

```lua
require('packer').startup(function(use)
    use 'wbthomason/packer.nvim'

    use 'Julian/lean.nvim'
    use 'neovim/nvim-lspconfig'
    use 'nvim-lua/plenary.nvim'

    -- Optional Dependencies:

    use 'hrsh7th/nvim-cmp'        -- For LSP completion
    use 'hrsh7th/vim-vsnip'       -- For snippets
    use 'andrewradev/switch.vim'  -- For Lean switch support
    use 'tomtom/tcomment_vim'     -- For commenting motions
end)

require('lean').setup{
    abbreviations = { builtin = true },
    -- lsp = { on_attach = on_attach },
    -- lsp3 = { on_attach = on_attach },
    mappings = true,
}
```

4e554c4c:
This fixes a crash that occurs when lean is configured outside of a
`config` function from packer. Technically---instead of communicating
changes to the filetype configuration through a shared configuration
global, the relevant configuration (default filetype) is shared through
one single global.
Furthermore, if this global is written by the user before the plugin is
loaded, then this global will not be managed by the plugin. This allows
a user to configure default filetype even if the plugin is not loaded!

Fixes #275

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,6 +1,6 @@
 globals = {
     "_clean_buffer_count",
     "vim",
-    "lean_nvim_ft_options",
+    "lean_nvim_default_filetype",
     assert = {fields = {"message"}},
 }
--- a/ftdetect/lean.lua
+++ b/ftdetect/lean.lua
@@ -8,24 +8,17 @@ local find_project_root = require('lspconfig.util').root_pattern(
   'lean-toolchain'
 )
 
-lean_nvim_ft_options = {
-  default = "lean",
-  nomodifiable = {
-    '.*/src/lean/.*',       -- Lean 4 standard library
-    '.*/lib/lean/src/.*',   -- Lean 4 legacy standard library
-    '.*/lean_packages/.*',  -- Lean 4 dependencies
-    _LEAN3_STANDARD_LIBRARY .. '.*',
-    '/_target/.*/.*.lean'   -- Lean 3 dependencies
-  }
-}
-
 local function detect(filename)
   if filename:match('^fugitive://.*') then
     filename = pcall(vim.fn.FugitiveReal, filename)
   end
 
   local abspath = vim.fn.fnamemodify(filename, ":p")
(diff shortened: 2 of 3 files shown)


.luacheckrc
```diff
@@ -1,6 +1,6 @@
 globals = {
     "_clean_buffer_count",
     "vim",
-    "lean_nvim_ft_options",
+    "lean_nvim_default_filetype",
```
4e554c4c, line 4: would really like to document the behaviour of this global, but i think it would be good to do that in `doc/lean.text` and im not quite sure how to generate the help with lemmy-help (in particular what files are used for it)
It would probably be best to just merge this PR to fix the bug, and then I'll send another PR for a help overhaul

Julian:
Thanks! Will merge to fix the breakage but if you think of a way we could reproduce this in CI that'd probably be nice. I suspect it may be hard given we're at the mercy of however plenary sequences running files though.

Issue #145, opened 2021-10-04
Make calling require'lean'.setup{} idempotent


Julian:
Users occasionally will put this in a file they reload or re-source, so it'd be nice if it were able to be run multiple times.

Right now most of the state issues are simply with initializing global configuration, so this may not be too difficult to do.

(See #132 for some context.)

Relates to #39.

4e554c4c:
the solution in my config
```lua
local success, err = pcall(function ()
  require('lean').setup{}
end)
if not success then
  print(err)
end
```

Issue #39, opened 2021-06-05
Complain about unknown configuration options


Julian:
Right now an unknown option is silently ignored, which allows typos (or mistakes due to backward incompatibility).

PR #277, opened 2022-09-06
Fix idempotency issue

4e554c4c:
It turns out that most of these come from the `_DEFAULTS` pattern. This
is because of the following:
original state:
```lua
tbl = { _DEFAULTS = ... }
```
after configuration with empty table, `tbl = vim.tbl_extend(tbl._DEFAULTS, {})`
which practically replaces `tbl` by `tbl._DEFAULTS`:
```lua
tbl = { ... }
```
after this, the next call to `vim.tbl_extend(tbl._DEFAULTS, {})` fails, as
`tbl._DEFAULTS` is nil.

The solution to this is to just call `vim.tbl_extend(tbl, ...)` and store the
defaults in `tbl` itself. The `._DEFAULTS` field can be preserved by copying to
it after initialization.

Also `command` has been replaced with `command!` in vimscript and
`autocmd`s have been wrapped in an `augroup` which resets its state each
call. This is also common practice that avoids duplicate autocommands.

Fixes https://github.com/Julian/lean.nvim/issues/145

### DIFF
--- a/lua/lean/commands.lua
+++ b/lua/lean/commands.lua
@@ -116,11 +116,11 @@ end
 
 function commands.enable()
   vim.cmd[[
-    command LeanPlainGoal :lua require'lean.commands'.show_goal(false)
-    command LeanPlainTermGoal :lua require'lean.commands'.show_term_goal(false)
-    command LeanGoal :lua require'lean.commands'.show_goal()
-    command LeanTermGoal :lua require'lean.commands'.show_term_goal()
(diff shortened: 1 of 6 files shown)


Julian:
Thanks. This looks pretty reasonable. I'm not really happy with the way we have config now in some other ways which is #39 but this definitely looks like an improvement! Let me know when/if it's ready to review.

4e554c4c:
should be good to review now, thanks :)

Julian:
Looks great! Really appreciated.

Issue #240, opened 2022-03-08
Set the infoview cursor to the goal when updating


Julian:
Right now if the infoview contents are too long, we show the top lines.

It seems more natural perhaps to show the goal instead, which is almost the bottom modulo us showing diagnostics below the goal state.

Comments welcome if someone prefers the current behavior.

gebner:
I certainly prefer the current behavior.  Scrolling to the bottom of the diagnostics is rarely what I want.

> It seems more natural perhaps to show the goal instead, which is almost the bottom modulo us showing diagnostics below the goal state.

This is a reasonable compromise.  The full goal should still be shown though if it fits on the screen.  (That is, don't show the last line of the goal as the first line in the infoview.)

Julian:
>  Scrolling to the bottom of the diagnostics is rarely what I want.

The goal is what I was really suggesting as in the body yeah. Essentially `vim.fn.search('^⊢ ', 'c'); vim.cmd('normal z-')` -- i.e. "find first goal, then run z-". Do you still prefer the current behavior to that?

gebner:
Oh sorry, that wasn't really clear to me in the first post.  Scrolling to the ⊢ is fine.

PR #279, opened 2022-10-16
Set the cursor to the goal when updating infoview windows.

Julian:
The main motivation for this is that it makes the goal visible when the infoview is full of many lines (e.g. there are many hypotheses). In particular this happens a lot on vertical displays with horizontal infoviews (that don't offer many lines).

Closes: #240

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -819,7 +819,11 @@ function Pin:__started_loading()
 end
 
 function Pin:async_update(force)
-  if not force and self.paused then return end
+  -- FIXME: For one, we're guarding here against the infoview being updated
+  --        while it's closed, which if we continued, would end up calling
+  --        render. That doesn't seem right, somewhere that should happen
+  --        higher up than here.
+  if not force and self.paused or not self.__info.__infoview.window then return end
 
   local tick = self.__ticker:lock()
 
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -590,6 +590,16 @@ function Info:render()
   self.__renderer:render()
   if self.__diff_pin then self.__diff_renderer:render() end
 
+  -- Set the cursor to the line with first goal (just after the marker).
+  for i, line in ipairs(vim.api.nvim_buf_get_lines(self.__renderer.buf, 0, -1, false)) do
+    if line:find("^⊢ ") then
+      vim.api.nvim_win_call(self.__infoview.window,
+        function() vim.cmd.normal(i .. 'z-2l') end
+      )
+      break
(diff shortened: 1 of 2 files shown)


lua/lean/infoview.lua
```diff
@@ -590,6 +590,16 @@ function Info:render()
...
 
+  -- Set the cursor to the line with first goal (just after the marker).
+  for i, line in ipairs(vim.api.nvim_buf_get_lines(self.__renderer.buf, 0, -1, false)) do
+    if line:find("^⊢ ") then
+      vim.api.nvim_win_call(self.__infoview.window,
+        function() vim.cmd.normal(i .. 'z-2l') end
+      )
+      break
+    end
+  end
```
rish987, line 601: Hmm, it's been a bit so I'd have to review some of our codebase again, but where all is `Info:render()` being called? I have a concern that this will move around the view while you're interacting with the infoview -- it should only happen when you "switch" infoviews (i.e. move somewhere else in the source), not within a single interactive infoview "session.

rish987: Also, we need to guard this with `if self.__infoview.window then ...`, otherwise we get error messages when the window is closed.

Julian: Thanks! You seem correct, and I never finished separating out Infoview / removing Info. It's definitely counterintuitive to me that `render` is called on closed infoviews, but it seems it indeed is... Will have a look and add a test.

rish987 (approved):
Thank you! I think there won't be any issues with aligning-while-interacting on the Lean 4 side since we were careful to avoid unnecessary calls of `Info:render`, but Lean 3 I think is a different story. I'm not in a position to test this right now, but I think for example if you have a really long infoview text, and say go to the top and change the filter, it'll end up aligning the goal again which would be confusing.

Overall, I think this alignment logic should happen at a higher level, but I'm not sure how to "hook" something like that onto the end of an async call, perhaps we'll need to write our own API for that. But to play it safe, could we make this configurable so that someone can disable it if they do have issues?

Julian:
Thanks! I have to check the aligning-while-interacting behavior still I was planning on doing so before either pinging or merging, will do it in a bit (for now been using this for a day or two of light leaning and haven't seen any other issues yet).

I agree the alignment should happen at a higher level, but I'm still somewhat unhappy with the whole path to rendering -- it seems to happen "bottom up" with Pins being responsible for rendering, when I think it should live up in infoviews, and Pins be somewhat behaviorless. But still needs more thought... (`Infoview:render` doesn't exist at the minute, maybe it did previously, but yeah it's something I was thinking about as well when looking at this)

PR #280, opened 2022-10-27
Make stderr window height configurable and fix autoscroll.

rish987:
I've been dealing with very long `repr` traces recently so this would be very nice to have, just wanted to make sure we agree on the option name/location.

Edit: ~~Also added an "autoscroll" to automatically go to the end of the trace when we get new output.~~ ~~Wait we already have this? But it wasn't working for me, hmm~~ Fixed.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -335,6 +335,8 @@ Full Configuration & Settings Information
       -- Redirect Lean's stderr messages somehwere (to a buffer by default)
       stderr = {
         enable = true,
+        -- height of the window
+        height = 5,
         -- a callback which will be called with (multi-line) stderr output
         -- e.g., use:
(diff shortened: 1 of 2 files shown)


Julian:
Feature seems reasonable to me, though yeah think maybe it'd be a bit more consistent to use the option name we use for the infoview (i.e. `width` and/or `height` and/or `horizontal_position` if we even support splitting `stderr` vertically, I forget).

rish987:
Good point, just renamed it to "height" (as it seems we don't support a vertical split atm). Also squeezed in a fix for the autoscroll functionality.

Julian:
Cool, haven't tested it out but lgtm if/when it works and passes CI

PR #292, opened 2023-03-25
fix normal[^!] bug

wyrmsVassal:
Use `normal!` instead of `normal` to avoid bugs caused by custom mappings.


Julian:
Thanks! We may be on new enough nvim now that we can avoid using vim.cmd entirely here I forget... but certainly better than before.

PR #307, opened 2023-09-23
separate lean 4 `sorry` fill into single goal and multiple goal alternatives

gihanmarasingha:
If there is a single remaining goal, the sorry fill code just adds a `sorry` in the next line. If there are multiple goals, a `· sorry` is added. Indentation is adjusted accordingly.

Tests are provided.


Julian:
Looks great!

PR #316, opened 2023-11-06
feat: initial loogle client

hargoniX:
This is the initial version I hacked together yesterday, I've never written code for a bigger Neovim plugin so I don't quite know what quality standard we have here and how we want things to look and work?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,6 +57,7 @@ or with `vim-plug <https://github.com/junegunn/vim-plug>`_:
     Plug 'hrsh7th/vim-vsnip'       " For snippets
     Plug 'andrewradev/switch.vim'  " For Lean switch support
     Plug 'tomtom/tcomment_vim'     " For commenting motions
+    Plug 'nvim-telescope/telescope.nvim' " For Loogle search
 
 ``lean.nvim`` already includes syntax highlighting and Lean filetype support, so installing the ``lean.vim`` (i.e. non-neovim) plugin is not required or recommended.
 
@@ -80,6 +81,8 @@ Features
 * Lean library search path access via ``lean.current_search_path()``, suitable for use with e.g. `telescope.nvim <https://github.com/nvim-telescope/telescope.nvim/>`_ for
   live grepping.
   See the wiki for `a sample configuration <https://github.com/Julian/lean.nvim/wiki/Configuring-&-Extending#live-grep>`_.
+
+* If `telescope.nvim <https://github.com/nvim-telescope/telescope.nvim>`__ is present a `:Telescope loogle` command is available as a frontend for the `Loogle <https://loogle.lean-lang.org>`_ JSON API.
 
 Configuration & Usage
 ---------------------
--- a/lua/telescope/_extensions/loogle/loogle_search.lua
+++ b/lua/telescope/_extensions/loogle/loogle_search.lua
@@ -1,0 +1,101 @@
+local pickers = require('telescope.pickers')
+local actions_state = require 'telescope.actions.state'
+local finders = require('telescope.finders')
+local conf = require('telescope.config').values
+local actions = require('telescope.actions')
+local previewers = require('telescope.previewers')
+
+local curl = require 'plenary.curl'
+
+local M = {}
+
+M.ask_for_type = function()
+  local type = ''
+  local on_confirm = function(input) type = input end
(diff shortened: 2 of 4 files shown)


Julian:
Very nice, this looks like a great (and probably near mergeable) start!

As for standards -- perhaps you could add a few docstrings to the extension module, but otherwise seems pretty good to me! We use LuaLS type annotations -- the docs for those are [here](https://luals.github.io/wiki/annotations/) or should be cargo cultable from other functions in the plugin that you see.

Also, the tests will pass if you make the telescope dependency optional, which you can do by wrapping the `require('telescope')` in `pcall` so that we only enable the functionality when telescope is installed. Let me know if that isn't clear enough and I can put a suggestion on the code.

I'll leave one or two other small comments as I read through the code but yeah thanks again for sending!

lua/telescope/_extensions/loogle/loogle_search.lua
```diff
@@ -0,0 +1,101 @@
...
+end
+
+M.find = function(telescope_opts)
+  telescope_opts = vim.tbl_extend('keep', telescope_opts or {}, {})
+
+  local type, results, err
+
+  type, err = M.ask_for_type()
+  if err then
+    print(err)
```
Julian, line 55: Probably instead of printing for these we should use `vim.ui.notify` with an `error` log level.

Julian (commented):
Some minor comments but yeah looks nice!

Other possible improvements might be to have an alternate action in the telescope picker which also adds the imports (or maybe we should do that by default), or which *only* adds the imports? As I say I haven't used Loogle yet other than just trying it via your extension :), what's your take?

lua/telescope/_extensions/loogle/loogle_search.lua
hargoniX: `vim.ui.notify` doesn't seem to exist on my Neovim, is this a nightly feature?

hargoniX:
> Other possible improvements might be to have an alternate action in the telescope picker which also adds the imports (or maybe we should do that by default), or which _only_ adds the imports? As I say I haven't used Loogle yet other than just trying it via your extension :), what's your take?

I don't think we should always add the import, we would have to determine if the thing is already imported (in particular transitive imports are a thing right now...) and in addition to that there might be some requirements on ordering etc. in the future. But I do agree it would be cool to have an additional action that gives you the thing + the import, I'll try to add that.

lua/telescope/_extensions/loogle/loogle_search.lua
Julian: It's `vim.notify`, sorry.

hargoniX:
And additional thing I was wondering: Can we make it such that you can type unicode down there? Otherwise lots of math lemmas are out of reach for the vim search.

Julian:
We do [this](https://github.com/Julian/lean.nvim/blob/main/lua/lean/abbreviations.lua#L250-L251) to attach 2 autocmd events when entering command mode windows (`q:` and `q/`) for abbreviations.

If telescope fires a similar autocmd we can do the same, or else you can do it "manually" by binding those same few functions I think (`convert` and `insert_char_pre`) inside the Telescope typing buffer.

Julian (approved):
Looks great!

There's extra stuff we can add, like adding `imports` or unicode abbreviation support for the input prompt, but as a first version it's perfect!

Can you just add a line to the README mentioning this exists (maybe next to the `current_search_paths` telescope note), and maybe a screenshot to the [wiki](https://github.com/Julian/lean.nvim/wiki/Configuring-&-Extending) as well?

After that LGTM, feel free to merge, you've now got the power.

README.rst
```diff
@@ -81,6 +82,9 @@ Features
   live grepping.
   See the wiki for `a sample configuration <https://github.com/Julian/lean.nvim/wiki/Configuring-&-Extending#live-grep>`_.
 
+* If `telescope.nvim <https://github.comnvim-telescope/telescope.nvim>`_ is present a `:Telescope loogle` command
```
Julian, line 85: Minor typos, and an extra `_` will make the nonsense cryptic reST error go away:

```suggestion
* If `telescope.nvim <https://github.com/nvim-telescope/telescope.nvim>`__ is present a `:Telescope loogle` command is available as a frontend for the `Loogle <https://loogle.lean-lang.org>`_ JSON API.
```

PR #318, opened 2023-11-06
Tweak the Loogle search implementation slightly in a few ways.

Julian:
* Fix Loogle sometimes returning responses with newlines, breaking nvim_buf_set_lines which wants none -- e.g. Complex.cos_add_mul_I returns a type with newlines in it. What we do now is split, then indent any newlines so the preview file looks "reasonableish"
  * Make the result be `put` charwise, so that e.g. #check <loogle here> will never insert extra lines
  * Make it usable outside of telescope via `require'lean.loogle'.search()`, which simply returns results
  * Add some more typing for Loogle API responses
  * Use Lua errors (exceptions) in a place or two to avoid some manual error checking

There's probably a few small other things that can be tidied.

### DIFF
--- a/lua/lean/loogle.lua
+++ b/lua/lean/loogle.lua
@@ -1,0 +1,48 @@
+local curl = require('plenary.curl')
+
+local loogle = {}
+
+---@class LoogleResult
+---@field name string
+---@field type string
+---@field module string
+---@field doc string
+
+---Search Loogle for the given type.
+---@param type string The type pattern to look for.
+---@return LoogleResult[]? results Loogle hits in the JSON API format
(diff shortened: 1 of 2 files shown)


Julian:
These are all quite minor I think, and mostly was a way to just poke at your code. Will give it a fresh look in the morning before merging, but @hargoniX if you have any comments feel free to lemme know!

Julian:
I meant to check if Loogle ever returns *names* with newlines for any absurdly long one. Hopefully not but leaving a just in case note to check.

Julian:
Thanks both!

Issue #333, opened 2024-03-15
Files opened under */src/lean/ are unmodifiable


mars0i:
This seemed pretty strange, but I have now figured out the source of the problem.

If I open any file named <name>.lean in neovim in a directory named "lean" under one named "src", and if I try to modify the file, I get nvim error "E21 Cannot make changes, 'modifiable' is off" until I run `:set modifiable`, regardless of file permissions.

I think this has to do with the `nomodifiable` configuration option.  If I add this line
```
 \ ft = { nomodifiable = {}, },
```
within
```
lua require('vfiler/config').setup{ options = {
...
}
```
the problem goes away.  This line in ft.lua seems to be the cause:
```
    '.*/src/lean/.*', -- Lean 4 standard library
```

This problem was a real pain to figure out.  I had no idea where to look until I did a lot of testing to find what I could and could not edit files.  It wasn't clear that lean.nvim was the culprit.

From my point if view, it's natural to put all of my source in a directory named `src`, and within that, I classify different kinds of source code by the name of the language.  So of course when Is started using Lean, I put my Lean source in `src/lean`.  This seems like a reasonable thing to me.  

I understand why it's reasonable to include a test for the standard library to prevent editing it by accident.  Would it be possible to add more complicated test to find the standard lib directory instead?   

If not, maybe the README could include a warning that certain directory path patterns will cause problems.  

(I'd be happy to add a PR for the README if you want.  I don't know enough about the standard library to change the path regex test, though.)

Thanks.   lean.nvim is great!  It's a real pleasure to use.  (At first I was working in a directory that wasn't under `src/lean`.)

Julian:
Sorry for the trouble.

> Would it be possible to add more complicated test to find the standard lib directory instead?

I don't think there is a simple one unfortunately. There's no other reliable way of knowing which directory is the standard library one that we could think of, which seemed still reasonably safe given that in Lean 4 it's discouraged to use `src` directories *within* packages. Of course your setup is precisely the one that would be a false positive (and seems otherwise fine certainly) -- but ultimately some false positive seems unavoidable when just looking at paths. And to be explicit -- new users who accidentally modify some core lean file ultimately ended up having huge troubles historically, so it seemed like a good tradeoff even if it has some potential for false positives.

[...]

PR #334, opened 2024-03-19
Added warning about E21 error due to nomodifiable.

mars0i:
Wasn't sure where to put the new text for #333.  It's now at the end of "Configuration & Usage". 

(Other options include:
- As a comment in the "Full Configuration & Settings Information" code block.
- In the wiki.)


Julian:
Cool! Let's try this. Thanks again for helping out, really appreciated!

PR #336, opened 2024-05-22
chore: adapt to changes in client fields for `vim.lsp.client`

mattrobball:
`get_language_id` has moved out from under `config` and up a level, see https://neovim.io/doc/user/lsp.html#lsp-client.


mattrobball:
`nvim --version` returns
```sh
NVIM v0.10.0
Build type: Release
LuaJIT 2.1.1713773202
```

Julian:
Thanks!

PR #337, opened 2024-05-25
feat(infoview): more buffer options

### LINKED ISSUE #332: Set winfixbuf when available (0.10+) on our infoview windows.
https://github.com/neovim/neovim/issues/12517

MithicSpirit:
Sets `nocursorcolumn`, `colorcolumn=""`, and `winfixbuf` (when available).
`'nocursorcolumn'` and `'colorcolumn'` are to prevent global options from
affecting the infoview buffer (the former is for consistency with
`'nocursorline'`, which is already present). `'winfixbuf'` is a new feature
in Neovim 0.10 that restricts the window to that particular buffer.

Closes #332.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+TODO*
+
 leanpkg.path
 *.olean
 .lake/
--- a/ftplugin/leaninfo.lua
+++ b/ftplugin/leaninfo.lua
@@ -1,9 +1,14 @@
 vim.bo.modifiable = false
 vim.bo.undolevels = -1
 vim.wo.cursorline = false
+vim.wo.cursorcolumn = false
+vim.wo.colorcolumn = ""
 vim.wo.number = false
 vim.wo.relativenumber = false
 vim.wo.spell = false
 vim.wo.winfixheight = true
 vim.wo.winfixwidth = true
 vim.wo.wrap = true
+if vim.fn.exists('&winfixbuf') > 0 then
+  vim.wo.winfixbuf = true
+end
--- a/lua/lean/abbreviations.lua
+++ b/lua/lean/abbreviations.lua
@@ -68,7 +68,7 @@ function abbreviations.show_reverse_lookup()
       end
     end
   end
-  vim.lsp.util.open_floating_preview(lines)
+  vim.lsp.util.open_floating_preview(lines, 'markdown')
 end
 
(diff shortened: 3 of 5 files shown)


Julian:
Thanks.

0.10 has `winfixbuf` which kind of makes `bufhidden` irrelevant (as it doesn't allow the buffer to be switched). #332 is a reminder to set it.

I think we also already set `buftype`? I haven't tried in an isolated setup so it's possible it's something I do, but my infoviews are certainly already set to that. Are yours not?

I also personally don't set `colorcolumn` for any window with `buftype=nofile`, but I think it's fine to explicitly not set colorcolumn, but then why not do that directly rather than via textwidth?

The place to set it (along with any others of these we do go with) would be [here alongside the other infoview options](https://github.com/Julian/lean.nvim/blob/main/ftplugin/leaninfo.lua).

MithicSpirit:
> 0.10 has `winfixbuf` which kind of makes `bufhidden` irrelevant (as it doesn't allow the buffer to be switched). #332 is a reminder to set it.

Good idea. Setting it now.

> I think we also already set `buftype`? I haven't tried in an isolated setup so it's possible it's something I do, but my infoviews are certainly already set to that. Are yours not?

They seem to be, but grepping for `buftype` finds no matches. I guess some function is setting automatically.

> I also personally don't set `colorcolumn` for any window with `buftype=nofile`, but I think it's fine to explicitly not set colorcolumn, but then why not do that directly rather than via textwidth?

I have `'colorcolumn'` and `'textwidth'` set globally, so it shows up independent of `'buftype'` (which is buffer-local rather than window-local anyway). I was using `'textwidth'` rather than `'colorcolumn'` because I was having issues setting window-local options (in hindsight, that's probably because of `opts`), but using the file you linked below it seems to work.

> The place to set it (along with any others of these we do go with) would be [here alongside the other infoview options](https://github.com/Julian/lean.nvim/blob/main/ftplugin/leaninfo.lua?rgh-link-date=2024-05-26T14%3A24%3A43Z).

Ah, I didn't see this. Yeah, this is a lot better.

MithicSpirit:
Hold on while I figure out why the `'winfixbuf'` check isn't working for the macOS build.

MithicSpirit:
Apparently 0 is truthy in Lua 🤦

PR #338, opened 2024-05-25
fix(abbr): don't stop abbr on `{}()`

MithicSpirit:
Since open braces/parentheses are almost always preceded by spaces, and
close braces/parentheses are almost always succeeded by spaces, there's
no real advantage to immediately converting an abbreviation when these
are seen. Moreover, this meant that the abbreviations `\{{` and `\}}`
were impossible to use (and we had to resort to `\c[` and `\c]`
instead). Overall, this change feels like a net positive (and users can
still override by just pressing `<Tab>` anyways.

This also makes a couple of minor changes with no noticeable effect
- Passing in `{syntax}` to `vim.lsp.util.open_floating_preview()`, which is
  documented as a mandatory parameter.
- Not returning `vim.schedule()` calls in `insert_char_pre()`, since
  returning truthy values from an autocmd callback causes the autocmd to
  be deleted. `vim.schedule()` is documented as always returning `nil`, so
  this doesn't really matter, but this is clearer.

### DIFF
--- a/lua/lean/abbreviations.lua
+++ b/lua/lean/abbreviations.lua
@@ -68,7 +68,7 @@ function abbreviations.show_reverse_lookup()
       end
     end
   end
-  vim.lsp.util.open_floating_preview(lines)
+  vim.lsp.util.open_floating_preview(lines, 'markdown')
 end
 
 local abbr_mark_ns = vim.api.nvim_create_namespace 'lean.abbreviations'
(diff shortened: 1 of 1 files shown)


Julian:
Thanks. I honestly don't recall why we special cased these, but this seems OK to me (and the tests pass after a rebase). Merging, appreciated.

PR #348, opened 2024-08-08
specify file extension for abbreviations

chenson2018:
This is my attempt to implement what was [suggested on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/lean.2Envim/near/456570614) for allowing the lean.nvim abbreviations for other file extensions.

The usage here would be that you call something like `require('lean.abbreviations').enable('*.tex', {})` to add a file extension. As I've not ever written much Lua or nvim plugins previously I'm not sure at all about what is considered idiomatic, so please give feedback as needed!

### DIFF
--- a/lua/lean/abbreviations.lua
+++ b/lua/lean/abbreviations.lua
@@ -256,7 +256,7 @@ function abbreviations.convert()
   vim.api.nvim_win_set_cursor(0, { row1 + 1, col1 + new_cursor_col_shift })
 end
 
-function abbreviations.enable(opts)
+function abbreviations.enable(pattern, opts)
   abbreviations.leader = opts.leader or '\\'
 
   abbreviations.abbreviations = abbreviations.load()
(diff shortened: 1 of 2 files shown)


Julian:
Awesome! Thanks a ton for giving this a shot. I'll take a closer look in a bit (likely tomorrow) -- at first glance it seems good! One additional thing you might try is to add a test to the [spec file](https://github.com/Julian/lean.nvim/blob/main/spec/abbreviations/abbreviations_spec.lua) which tests this functionality.

You should mostly be able to copy a test that's there and simply create a file with some other file extension.

In order to run the tests locally you'll need [just](https://github.com/casey/just?tab=readme-ov-file#installation) (`brew install just` on macOS if you're there) -- or otherwise you can just rely on GitHub actions running them, though it'll take a bit longer obviously. Ask if that's not specific enough help!

chenson2018:
I took a stab at the unit testing, making changes to the existing helper function `set_unique_name_so_we_always_have_a_separate_fake_file` to support different file extensions and copying over a test as suggested.

Julian:
Awesome! I did that test modification in a slightly different way, but now merging! Thanks again, and let me know how this goes!

chenson2018:
I saw your follow up commit to the testing, that's much cleaner. Thanks!

PR #357, opened 2024-10-22
Add a default mapping for `:LeanRestartFile` and related tips

utensil:
[Zulip](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/lean.2Envim/near/478285801)

- [x] Add a default mapping for `:LeanRestartFile`
  - [x] stylua 
  - [ ] test
- [x] Add related tips, restart file, restart server
  - [ ] review wordings

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,11 +139,15 @@ This can be configured by putting a line at the top of your `~/.config/nvim/init
 | `<LocalLeader>v`     | interactively configure infoview view options                       |
 | `<LocalLeader><Tab>` | jump into the infoview window associated with the current lean file |
 | `<LocalLeader>\\`    | show what abbreviation produces the symbol under the cursor         |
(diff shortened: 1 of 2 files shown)


Julian:
Thanks! The style warning is markdownlint being silly, we only care about GitHub's parser. The wordings actually look fine to me too -- if you do care to add a test it should be pretty trivial to do so although we don't actually test *all* the mappings we set really currently, we just pick some single one and make sure it's properly mapped  -- so I'm happy to merge this as-is and add a linter exception -- if you do have any other suggestions or improvements on this or anything though they're always welcome!

utensil:
Thanks, I'll try to figure out how to set up the test tomorrow, and add the test.

utensil:
Updated and tested locally from nvim, it seems to work!

PR #358, opened 2024-10-23
Add test for LeanRestartFile mapping

utensil:
This is the test for #357 . It passes in the sense that

```
just retest spec/mappings_spec.lua
```

returns 0, and I can manage to fail it too.

But it's mostly a dummy test, checking the key mapping is set, but I don't really know how to check if it can restart file.

So, feel free to close without merging.

Also, I can't pass all existing tests locally, some of them are failing without modifications, the log file is attached.

[just-test.log](https://github.com/user-attachments/files/17486662/just-test.log)

### DIFF
--- a/spec/mappings_spec.lua
+++ b/spec/mappings_spec.lua
@@ -34,6 +34,17 @@ describe('mappings', function()
         end
       end)
     )
+
+    it(
+      'can restart file',
+      clean_buffer(function()
+        local restart_file_lhs = '<LocalLeader>r'
+        local restart_file_rhs = '<Cmd>LeanRestartFile<CR>'
+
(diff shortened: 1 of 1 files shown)


Julian:
Thanks! Yeah writing a test for the actual restart functionality is a bit trickier, though should still be doable, it's been on my list of "known missing tests". No harm merging this I suppose.

> Also, I can't pass all existing tests locally, some of them are failing without modifications, the log file is attached.

Are you on nightly? There's some flakiness in the tests which seems to only manifest on nightly. If you run the suite a few times, does it pass any of them?

utensil:
Not really, I'm on nvim v 0.10.20.

PR #367, opened 2024-12-06
fix(keymap): not shadow keymap desc so make blink.cmp happy

tim3nd:
fix Saghen/blink.cmp#441

Currently blink.cmp use `desc` to identify blink.cmp keymap, so shadow `decs` option in  lean.nvim would make stack overflow in blink.cmp side.

I think this should mainly be handled for blink.cmp, but restore `desc` option after cleanup by lean.nvim seems also a reasonable default.

### DIFF
--- a/lua/lean/abbreviations.lua
+++ b/lua/lean/abbreviations.lua
@@ -164,6 +164,7 @@ local function insert_char_pre()
             expr = imap.expr,
             unique = imap.unique,
             callback = imap.callback,
+            desc = imap.desc,
           })
         end
       end


Julian:
Thanks, I've seen this error myself in playing with blink.cmp and had no idea why I couldn't see where it was being caused.

tim3nd:
Just FYI, blink.cmp use `desc` [here](https://github.com/Saghen/blink.cmp/blob/main/lua/blink/cmp/keymap/fallback.lua#L16) to deduplicate of the fallback keymap configuration.  :)
I also tried to improve this in blink.cmp but haven't found a clear and suitable method yet.

PR #373, opened 2025-01-08
fix goto_last_window

funemy:
Jumping back to the associated lean file from the current infoview does not work in the current version.

 The bug is that the `self` in `goto_last_window` refers to the metatable `Info` instead of the `new_info` table (where `last_window` is set).

I'm not familiar with this project nor Lua, so hopefully this fix makes sense.
Thanks for the awesome nvim plugin, btw!

### DIFF
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -589,24 +589,24 @@ end
 
 ---@return Info
 function Info:new(opts)
-  local pins_element = Element:new {
-    name = 'info',
-    events = {
-      goto_last_window = function()
-        if not self.last_window then
-          return
-        end
-        vim.api.nvim_set_current_win(self.last_window)
-      end,
-    },
-  }
   local new_info = setmetatable({
     pins = {},
     __infoview = opts.infoview,
-    __pins_element = pins_element,
+    __pins_element = Element:new { name = 'info' },
     __diff_pin_element = Element:new { name = 'diff' },
     __win_event_disable = false, -- FIXME: This too is really confusing
   }, self)
+
+  new_info.__pins_element.events = {
+    goto_last_window = function()
+      print(vim.inspect(new_info))
+      if not new_info.last_window then
+        return
+      end
+      vim.api.nvim_set_current_win(new_info.last_window)
+    end,
+  }
+
   new_info.pin = Pin:new {
     id = '1',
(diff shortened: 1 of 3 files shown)


Julian:
Thanks a lot for the PR and the kind words!

That's... "funny", you'd have thought I'd have noticed that but I guess I personally use the jumplist more often than this mapping.

It'd be good to add a test for this I think before merging. You might be able to cargo cult the tests you see e.g. in spec/infoview/open_close_spec.lua in order to create some simple new spec file without knowing much Lua (or much about the project yet). But I'm also quite happy to help or to do so myself if you run into trouble.

Does that sound like something you could have a look at? Regardless thanks for sending the PR!

funemy:
@Julian I added some tests for jumps between Lean file and infoview, lmk if they make sense.

lua/lean/infoview.lua
```diff
@@ -589,24 +589,24 @@ end
...
     __infoview = opts.infoview,
-    __pins_element = pins_element,
+    __pins_element = Element:new { name = 'info' },
     __diff_pin_element = Element:new { name = 'diff' },
     __win_event_disable = false, -- FIXME: This too is really confusing
   }, self)
+
+  new_info.__pins_element.events = {
+    goto_last_window = function()
+      print(vim.inspect(new_info))
```
Julian, line 602: Can you kill this `print`? Other than that this looks ready to go, so yeah thanks again!

spec/infoview/jump_spec.lua
```diff
@@ -0,0 +1,33 @@
...
+    lean_window = vim.api.nvim_get_current_win()
+
+    vim.cmd('edit! ' .. fixtures.project.some_existing_file)
+    local current_infoview = infoview.get_current_infoview()
+
+    current_infoview:open()
+    -- Both Lean and infoview windows exist
+    assert.windows.are(lean_window, current_infoview.window)
+
+    vim.cmd 'LeanGotoInfoview'
```
Julian, line 25:
```suggestion
    vim.cmd.LeanGotoInfoview()
```

should work here FWIW.

lua/lean/infoview.lua
funemy: oops, forgot to remove this, it's gone now.

PR #385, opened 2025-04-05
Remove some keywords in leanDeclaration

stephen-huan:
In lean4, the following seem to be no longer keywords. Since lean3 has been removed (3c0b48374146abd651da5623f3f73f96121a616f), drop them as well.

```lean
def lemma := 1
def definition := 1
def constant := 1
def abbreviation := 1
```

cf. `def theorem := 1`, which gives the error `unexpected token 'def'; expected identifier` as expected.


stephen-huan:
Test failures seem unrelated.

Julian:
`lemma` is re-added by Mathlib as a keyword.

The other 3 seem reasonable to remove though.

Julian:
Thanks!

PR #394, opened 2025-05-24
Add config for goals accomplished and no goals

stephen-huan:
Purely vanity change, feel free to close if you think it's unnecessary!

Inspired by the goal completion message

> After using [`rfl`](https://lean-lang.org/doc/reference/latest//Tactic-Proofs/Tactic-Reference/#rfl), the resulting state is:
> All goals completed! 🐙

in the [lean language reference](https://lean-lang.org/doc/reference/latest//Introduction/#code-samples). I was curious where this came from, and bisected it to [leanprover/verso](https://github.com/leanprover/verso/blob/41b85d429fcdb33349115edc063078dd98ee7c0b/src/verso-blog/Verso/Genre/Blog/Template.lean#L138) ([41b85d4](https://github.com/leanprover/verso/commit/41b85d429fcdb33349115edc063078dd98ee7c0b)).

Technically the message for no goals can already be configured with `vim.g.lean_no_goals_message`, but this is an implementation detail and must be done with an autocmd like the following as the lsp itself sets it on initialization.

```lua
vim.api.nvim_create_autocmd("LspAttach", {
    group = "vimrc",
    pattern = { "*.lean" },
    callback = function()
        vim.g.lean_no_goals_message = "No goals!"
    end,
})
```

### DIFF
--- a/lsp/leanls.lua
+++ b/lsp/leanls.lua
@@ -217,15 +217,16 @@ return {
   },
   on_init = function(_, response)
     local version = response.serverInfo.version
+    local markers = CONFIG.goal_markers
     ---Lean 4.19 introduces silent diagnostics, which we use to differentiate
-    ---between "No goals." and "Goals accomplished. For older versions, we
+    ---between "No goals." and "Goals accomplished". For older versions, we
     ---always say the latter (which is consistent with `lean.nvim`'s historic
     ---behavior, albeit not with VSCode's).
     ---
     ---Technically this being a global is wrong, and will mean we start
     ---showing the wrong message if someone opens an older Lean buffer in the
     ---same session as a newer one...
-    vim.g.lean_no_goals_message = vim.version.ge(version, '0.3.0') and 'No goals.'
-      or 'Goals accomplished 🎉'
+    vim.g.lean_no_goals_message = vim.version.ge(version, '0.3.0') and markers.no_goals
+      or markers.goals_accomplished
   end,
 }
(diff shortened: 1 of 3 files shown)


Julian:
Thanks!

> Technically the message for no goals can already be configured with vim.g.lean_no_goals_message but this is an implementation detail

Yeah this indeed was just meant as an implementation detail (and really just because of <4.19) -- though I did have this exact customization functionality on my TODO list (albeit quite low priority). So I'm happy to take a PR for it (thanks!) though what I had in mind originally was to allow specifying a table where the key is the number of goals and the value is the message (in other words e.g. `lean.infoview.goals_message[n]` should return the message for `n` goals, and so you'd override 0 if you want to change the goals accomplished message.

I'm not 100% bothered though with whatever configuration so I'm fine with just doing it as you have it -- though I would like a test. Could you try to have a look at that, you should be able to essentially cargo cult an existing spec file for this.

> The macos tests seem a bit flaky, but I guess they're re-ran until they succeed?

Yeah I've put in a lot of work to try to get the test suite to be less flaky but there's still a bit of missing synchronization. If/when I find where it is I'll fix it, but as I say I've done a ton of that already to get us to this point where it at least passes *most* of the time. If you do see something specific that's missing (usually a missing `wait`) of course let me know.

stephen-huan:
Thanks for working on this... very sorry for dropping the ball on this, didn't get around to it in time. Thanks!

PR #403, opened 2025-06-27
Add `orientation` opt to force infoview orentation

lucianchauvin:
I have found that for some projects, it would be nice to force a horizontal info view in the case of long lines in the `.lean` file and the infoview.


Julian:
Thanks. This seems reasonable -- could you try to add a test by following one of the existing files, say [this one](https://github.com/Julian/lean.nvim/blob/main/spec/infoview/layout/portrait_spec.lua)? CI is a bit flaky at the moment and sometimes requires running multiple times (as will `just test` if you're on macOS) because there's a race condition in one of the tests which depends on Mathlib, but it should hopefully still not be too difficult to do, and I'm happy to help.

lucianchauvin:
The current infoview tests should be enough to make sure this isn't breaking current functionality. I can add tests to make sure it's forcing.

Issue #427, opened 2025-09-18
Usage of deprecated `nvim-lspconfig` framework


iamrecursion:
My config only depends on `nvim-lspconfig` for `lean.nvim`, and I tried to update my plugins today and got pinged by [this change](https://github.com/neovim/nvim-lspconfig/commit/1f7fbc34e6420476142b5cc85e9bee52717540fb). You are probably already aware, but I figured I would raise it just in case!

StamesJames:
It seems, that the `leanls.lua` has not been migrated to the new system yet. It is in `lua/lspconfig/configs/` but not in `lsp/` 
I found this related issue https://github.com/neovim/nvim-lspconfig/issues/3705

StamesJames:
I tried to simply copy the old default configs to the plugins, but this does not work for some reason.

michael-novak-math:
I now don't get any error message anymore! I'm not sure if that's because of a recent update to the plugin or because I created a lean project properly now, but anyway it seems like the problem is solved, at least for me.

PR #428, opened 2025-09-20
Fix deprecated use of lspconfig

StamesJames:
I saw that `leanls.lua` is not yet in the new `lsp/` directory in the `lspconfig` plugin. I wrote the PR [4084](https://github.com/neovim/nvim-lspconfig/pull/4084) to the `lspconfig` repo adding it. Once some default configs for leanls are in `lsp/` this small change should work without the message that the deprecated framework is used. 
This is related to Issue #427

### DIFF
--- a/lua/lean/lsp.lua
+++ b/lua/lean/lsp.lua
@@ -225,7 +225,8 @@ function lsp.enable(opts)
         or 'Goals accomplished 🎉'
     end,
   })
-  require('lspconfig').leanls.setup(opts)
+  vim.lsp.enable 'leanls'
+  vim.lsp.config('leanls', opts)
 end
 
 ---Restart the Lean server for an open Lean 4 file.


Julian:
Thanks. Please either rebase or push an empty commit so we can see if CI agrees.

Julian:
I hadn't seen your push but it seems CI here says this still isn't quite right, there's lots of failures. If you pull main you'll at least have the 0.10 failures go away as like I mentioned we'd have to drop 0.10 to do this (which I've now done), but the 0.11 ones still seem "real".

StamesJames:
I will look into it. Maybe I forgot something in the lspconfig PR and the configuration.

StamesJames:
I tested some of the failing test by hand and there they seem to work. Is it possible, that the automated tests do not use the nvim-lspconfig version where the leanls configs are already there?

Julian:
The test runs will use an isolated config, yes (one that generally pulls the latest versions of plugins when you run it).

The issues here will be more subtle, from what I saw they have to do with the fact that the change here is not a drop in replacement -- clients are attached at different times via the new interface than via the old one, either because of different autocmd events or sequencing.

StamesJames:
Yes, I also think it is something like this, because when I use the new version outside the tests, it works just fine. I will look into the tests and how they are implemented and try to figure out if the new systems attaches LSPs differently, so that the tests do not recognize it.

yd278:
Hi, I tried to modify these lua scripts myself and find a possible workaround:

In `leanls.lua` of the nvim-lspconfig, the `cmd` is set as a function. However, I received notifications saying that the `config` parameter is nil, and the ls processes is not started at all.

Therefore, I simply change it into   `cmd = { 'lake', 'serve' },` and use the new api
```
vim.lsp.config('leanls', opts)
vim.lsp.enable 'leanls'
```
I also swapped these two line to make sure the configs are set before it get enabled. Now the ls process can get started as normal.

Hope that would be helpful!

StamesJames:
Thank you, @yd278, that is very helpful. That your config parameter is nil is strange, when I run it, it is not.
According to the nvim help page, returning a function for `cmd` is supported
```
      • {cmd}                  (`string[]|fun(dispatchers: vim.lsp.rpc.Dispatchers, config: vim.lsp.ClientConfig): vim.lsp.rpc.PublicClient`)
```

When I wrote the config in lspconfig I wrote `cmd` as a function because otherwise the `root_dir` was not getting passed to the lsp properly.
Does your LSP register the Project you opened correctly?
Maybe the call to `vim.lsp.rpc.start` behaves slightly different from what is done when `cmd` is returned as `{"lake", "serve"}`

yd278:
Hi, @StamesJames , frankly speaking ,it doesn't. I thought it work because I open the file directly under the root directory. However, I tried a minimal setup with only:

    lazy.nvim
    lean.nvim
    nvim-lspconfig
    plenary.nvim

  and I can confirm that no lean/lake processes were started, a debug message `attempted to index local 'config' (a nil value)` appeared as well

Maybe some other plugins did something on your machine?

StamesJames:
I tried some things in the same settings the tests run with `just nvim` I found that the root dir still is not right, but that can be fixed by changing the `cmd` to
```lua
  cmd = function(dispatchers, config)
    local local_cmd = { 'lake', 'serve', '--' }
    return vim.lsp.rpc.start(local_cmd, dispatchers, { cwd = config.root_dir })
  end,
```
This also lets more tests succeed, but still some fail. 

One thing I noticed is, that when I open the project folder with `just nvim ./spec/fixtures/projects/Example/` and then open the `Example.lean` file everything works but when I directly open `just nvim ./spec/fixtures/projects/Example/Example.lean` no LSP is attached.

Julian:
Thanks for the work here folks -- I didn't follow up as I didn't actually have any further info to add, but after giving this a bit of time, the key I think was finding that vim.lsp.start no longer does [this cwd setting](https://github.com/neovim/nvim-lspconfig/blob/64e0a1c6ba39417ee81a893ebbb4f8e4a949445c/lua/lspconfig/manager.lua#L118-L120) that nvim-lspconfig did.

I have no energy really to see whether that was discussed upstream or not but I suspect it was an intentional behavior change (and one that probably just isn't well documented).

I've cherry picked a commit from here and then added on top what ended up fixing the test suite in 6a5c8d22824b0c154596a03e17df7c14290859c7 and 4f62922d43d87e682727a4d584c6e1fc39cace62.

The latter does more than just fix this issue, it also does some minor restructuring of where the LSP code lives, but as I say, the key was the cwd bit, so if anyone's extra curious you can try playing with that commit.

I'm going to close this now that it's been partially merged. Any feedback or further fixes are of course welcome.

yd278:
Hi, the new version doesn't work for me as I still need nvim-lspconfig for other languages, therefore the plugin will still go with leanls.lua in the lspconfig's folder.

Moreover, I tried a minimal setup with only lazy.nvim and newest lean.nvim, still experienced `attempted to index local 'config' (a nil value)` issue. No lean/lake process found in activity monitor. I also noticed that the macos test has failed. Could that be something related to the OS?

Vinh-CHUC:
I also experience this, on Linux in my case

https://github.com/Julian/lean.nvim/issues/444

PR #443, opened 2025-10-20
README.md: fix syntax error in example lazy.nvim plugin installation

cdunham:
NVIM v0.11.4 (via neovim-snap on Ubuntu 24.04) and the latest lazy.nvim gives the error "unexpected symbol near '{'" without the return keyword.


cdunham:
Makes sense, thanks.

PR #474, opened 2026-02-07
In infoview, avoid computing #element.text when it might be vim.NIL

scott7z:
This fixes tui.lua to avoid computing #element.text in some cases when
    it might be vim.NIL, which is an error.  In particular, CR on a symbol
    in the infoview window will exercise this problem.

    A new unit test checks that we no longer draw an error, and that the
    right content is presented.

    To reproduce by hand, edit
      example (h: ∃ a:Nat, a = 3) := by apply h
    and then CR on `a` in the info view.

    Test with
      just retest spec/infoview/tooltips_spec.lua


Julian:
Do you also see the test passing even without your change? In my case your reproduction steps don't reproduce the issue. They do for you?

scott7z:
> Do you also see the test passing even without your change? In my case your reproduction steps don't reproduce the issue. They do for you?

I added a new test specifically to exercise the failures I was seeing.  The pre-existing tooltips_spec didn't exercise the code path that was breaking, so it passed.  

However, I just remembered that I'm running nvim-nightly.  Sure enough, it doesn't reproduce using nvim-0.11.6.   Sorry for not checking that first!

scott7z:
I don't know how to provoke it to behave that way (except for the first case (doc==nil), which I was hitting all the time), and in any event, the language server can change the details of its behavior out from under us at any time.   

I think we want to check both what it does send, and also whatever it could send, at least in terms of optional fields in the protocol. The mock is simply to generate those (possibly artificial) responses so we do exercise the code paths on our end that are effected.

I guess you could argue that this is mixing end-to-end tests with unit tests, but putting it here seems parsimonious.

scott7z:
Is there anything else I can do to move this along?

Julian:
Thanks for the fix (and patience!)

Issue #257, opened 2022-05-03
Make how and when the infoview updates more configurable, particularly for cases when it goes blank


Julian:
Putting this here off the back of / in order to track [this Zulip request](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/lean.2Envim/near/280738234).

I want to be able to configure Infoview updates to be less "jittery" (and possibly then the default should be reconsidered). Often one can be typing things and the infoview will change to be empty, or say processing if that message is enabled, when it would be a lot more useful to continue showing the last infoview contents until the new ones are available.

It occurs to me that it may be cohesive if we introduce the same orange progress bar concept which we use in the sign column for Lean buffers to the *infoview* window, and thereby indicate "what's here isn't current anymore" not always by removing it, but by instead the infoview window's sign column.

Other ideas are of course welcome, I don't know that I have a perfect one yet, but I know I've been frustrated by similar things that it seems Bartosz is.

PR #485, opened 2026-03-16
Preserve stale goal state while processing

### LINKED ISSUE #369: Still show last compiled goal-state while `Processing file...`
When I'm writing a proof, I would love to be able to have a consistent point of reference of existing hypothesis in the scope. Currently on every justification step I have to stop, wait for the file to process, and then in 90% of cases look for a hypothesis that was already available to find its name and exact statement.

It would be so much faster to have an option (or have it be default) of seeing the previously compiled state with a small indicator of `Processing file...` at the bottom or to the side, which would allow to write proofs without needless interrupts.

valeratrades:
## Summary
When Lean is processing a file, the infoview now preserves the last rendered goal state instead of clearing it and showing only "Processing file...". This lets you reference existing hypotheses while writing proofs without waiting for recompilation.

- Adds `keep_stale_goal` option (default: `true`) to infoview config
- When enabled and Lean is processing, the previous goal state is kept visible
- The infoview window gets a `leanInfoProcessing` winhighlight (`NormalNC`) to visually indicate the displayed state is stale
- The stderr window is explicitly excluded from the processing tint via `NormalNC:Normal` override
- On first file open (no previous state), falls back to showing "Processing file..." as before
- Only preserves stale state when cursor is inside a declaration — moving to a comment between theorems correctly clears the infoview (upstream `contents_for` returned `nil` via nonexistent `components.EMPTY` for no-info positions, which our stale logic mistook for "processing")
- Fixes a pre-existing flaky test in `core_spec.lua` where the infoview cursor position persisted between tests causing `helpers.search` to fail
- Fixes a nightly-only test failure in `window_spec.lua` where neovim added a `style` field to `nvim_win_get_config()` output

### Demo
https://github.com/user-attachments/assets/e9e96fd8-746a-4442-9eea-38b07d3e960d

## Related issues
- Closes #369 — Still show last compiled goal-state while `Processing file...`
- Addresses #257 — Make how and when the infoview updates more configurable

## Try Yourself
if you wanna try yourself before merging, - here is the list of expected behavior and possible pitfalls.
- Open a Lean file, verify "Processing file..." shows on first load
- Edit a proof, verify the previous goal state persists while processing
- Verify the infoview background tints when showing stale state (unfocused)
- Verify the stderr window does not get tinted
- Move cursor to a comment between declarations, verify stale goal clears
- Set `keep_stale_goal = false` and verify original behavior is preserved
- Check switching to a different file does not show old one's goalstate while loading

### DIFF
--- a/lua/lean/config.lua
+++ b/lua/lean/config.lua
@@ -46,6 +46,7 @@
 ---@class lean.infoview.Config
 ---@field mappings? { [string]: ElementEvent }
 ---@field orientation? "auto"|"vertical"|"horizontal" what orientation to use for opened infoviews
+---@field keep_stale_goal? boolean preserve the last goal state while Lean is processing instead of clearing it
 ---@field view_options? InfoviewViewOptions
 ---@field severity_markers? table<lsp.DiagnosticSeverity, string> characters to use for denoting diagnostic severity
 
--- a/lua/lean/infoview.lua
+++ b/lua/lean/infoview.lua
@@ -1140,7 +1140,7 @@ local function contents_for(params, use_widgets)
     elseif options.show_no_info_message then
       return components.NO_INFO
     else
-      return components.EMPTY
+      return Element.EMPTY
     end
   end
 
@@ -1201,6 +1201,8 @@ Pin.update = a.void(function(self)
   local new_element = contents_for(self.__position_params, self.__use_widgets)
   if new_element == nil then
     local has_stale = self.__data_element ~= Element.EMPTY
+      and self.__data_element ~= components.NO_INFO
+      and self.__data_element ~= components.PROCESSING
       and self.__data_element_uri == current_uri
     if has_stale then
       -- Preserve stale goal, indicate staleness visually.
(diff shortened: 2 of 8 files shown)


Julian:
So I had a bit more of a think here and I think what was bothering me a bit previously -- we have to think about the `show_processing` option Specifically, that option currently decides whether we show "Processing..." or an empty infoview while processing. This change kind of makes its behavior not make sense any more, so at very least we should remove it -- but we could also *preserve* it (and not need to add another boolean flag like `keep_stale_goal`) by deprecating that option and introducing an `on_processing` hook which takes the previous render's contents and returns the new contents -- i.e. something like

```
function on_processing(previous)
  return previous
end
```

would be the default as you have here, but then

```
function on_processing(previous)
  return ""
end
```

would be `show_processing = false`, and `return "Processing..."` would be `show_processing = true` (there's a minor consideration there about whether we want to expose `Element` rather than asking users to return strings...)

I'm happy to take some of that after merging this given I let it sit here (sorry!) -- let me know though if you're willing to think that last bit through yourself.

(I also had a bit of a big refactor of pins -- in the course of deleting an object called `Info` that we used to have and now do not, which has caused a bunch of merge conflicts, though I don't think resolving them should be too difficult. Apologies!)

PR #495, opened 2026-04-06
Highlight ≠ and ^ as operators

antinomie8:
Thanks for the plugin !
The usual operators `+*-/` are highlighted as `leanOp` in `syntax/lean.vim`, but not `^` (exponentiation), and `=<>` are also highlighted as `leanOp` but not `≠`, so I suggest adding these.


Julian:
Thanks.

PR #510, opened 2026-05-07
Reduce the huge delay while typing by using conflated channel.

Neuron-Group:


Julian:
Hi there, when reporting an issue please include some sort of description of what you're reporting. (Ideally before a pull request).

Neuron-Group:
I got optimize the delay while typing by modify the update handle and build a conflated channel. It seems work speedily on my manual test, but I still need pass the full check.

Julian:
I have no idea what that means I'm afraid, please explain in a bit more detail, with a reproducer and/or screen recording when you file an issue.

PR #512, opened 2026-05-17
fix:skip kitty graphics probe inside tmux to avoid title corruption.

BitPusher16:
Fix suggested by claude. Works in my setup. Alacritty + Tmux + Nvim.

### DIFF
--- a/lua/kitty.lua
+++ b/lua/kitty.lua
@@ -58,7 +58,6 @@ local function probe(on_detected)
     end,
   })
   -- Query: transmit a 1x1 transparent pixel with a=q (query).
-  --vim.api.nvim_chan_send(2, string.format('\x1b_Gi=%d,s=1,v=1,a=q,t=d,f=24;AAAA\x1b\\', QUERY_ID))
   send_raw(string.format('\x1b_Gi=%d,s=1,v=1,a=q,t=d,f=24;AAAA\x1b\\', QUERY_ID))
 end
 
--- a/lua/kitty.lua
+++ b/lua/kitty.lua
@@ -25,6 +25,16 @@ end
 
 local QUERY_ID = 99999
 
+local function send_raw(data)
+  if vim.env.TMUX then
+    -- Wrap in tmux DCS passthrough; inner ESCs must be doubled
+    local escaped = data:gsub('\x1b', '\x1b\x1b')
+    vim.api.nvim_chan_send(2, '\x1bPtmux;' .. escaped .. '\x1b\\')
+  else
+    vim.api.nvim_chan_send(2, data)
+  end
+end
+
 ---Send a Kitty graphics protocol query to probe for support.
 ---The terminal responds with an APC sequence if it supports the protocol.
 ---Requires Neovim to forward APC responses via TermResponse (recent feature).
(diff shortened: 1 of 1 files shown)


lua/kitty.lua
```diff
@@ -48,7 +58,8 @@ local function probe(on_detected)
     end,
   })
   -- Query: transmit a 1x1 transparent pixel with a=q (query).
-  vim.api.nvim_chan_send(2, string.format('\x1b_Gi=%d,s=1,v=1,a=q,t=d,f=24;AAAA\x1b\\', QUERY_ID))
+  --vim.api.nvim_chan_send(2, string.format('\x1b_Gi=%d,s=1,v=1,a=q,t=d,f=24;AAAA\x1b\\', QUERY_ID))
```
Julian: please remove

BitPusher16:
Noting that I misnamed the first commit. This change doesn't skip the probe, it correctly wraps it.

Julian:
Thanks!

SanchayanMaity:
This introduces a bug when using `ghostty` + `neovim` + `tmux` where in any newly opened file, it adds a `=99999;OK` string on the last line.

BitPusher16:
Pasting below feedback from Claude. Seems we can fix one of two ways. Modify tmux check to also check for and require Alacritty, or modify tmux check to exclude ghostty. From Claude's comments below, seems that ghostty may be the outlier, so better to just skip the wrap if ghostty in use?

=== Claude ===

Ghostty supports the Kitty graphics protocol and also supports passthrough of APC sequences through tmux — unlike most terminals. So when your fix skips the probe inside tmux unconditionally, Ghostty users lose graphics support even though their setup would have worked fine.
The detect_from_env() function already handles this partially — it checks TERM_PROGRAM=ghostty and returns true. But if Fix 2 (skipping the probe) was used, the issue is that detect_from_env() may not fire in all cases inside tmux, because tmux typically overrides TERM and may not propagate TERM_PROGRAM from the outer terminal into the shell.
The proper fix is to not skip the probe for Ghostty, which means instead of blanket-guarding on TMUX, you should:

Still run the probe inside tmux, but wrap it with the DCS passthrough (Fix 1)
Or, only skip the probe if inside tmux and the terminal isn't known to support passthrough (i.e., not Ghostty/WezTerm)

Fix 1 (the DCS passthrough wrapping) was always the more correct solution — it lets terminals that support APC passthrough through tmux continue to work, while preventing the raw APC bytes from leaking into tmux's title handling on terminals that don't.

Julian:
Ideally would be nice to have someone on that setup try it to be sure it works. But yes detecting we're in ghostty sounds marginally better I think as you say.

BitPusher16:
Silly question, but is the nvim_chan_send (currently wrapped in send_raw()) at the end of probe() actually doing anything here? Seems it could just be removed. Maybe at one point it was planned to use the value returned from nvim_chan_send to indicate whether kitty is supported, but as written, the return value is just lost. It is not returned or otherwise used.

I missed this initially bc not being a lua developer I just thought the function was returning value of last expression, but lua apparently doesn't do that.

EDIT: Ah wait, sorry. I see the QUERY_ID is used in the callback. My mistake.

SanchayanMaity:
I think Claude only made matters worse. I am not an expert but here is what I found after digging in.

If I use debug `print`, after sending that raw sequence, `ev.data` seems to be `"\27[?1;2;4c"` and not even the kitty APC response `\x1b_Gi=99999;OK\x1b\\`. That is the terminal's DA1 (Device Attributes 1) response IIUC and the kitty APC response seems to just leak through. Neovim's `TermResponse` auto command is never triggered. With `tmux`, it seems one can't realistically use pass through for anything that requires a response. Found a related commit in neovim, see `0c0352783f`.

I think in `tmux`, the probe should just be skipped. From my testing, it also seems that `graphics = { enabled = false }` does not really work and `kitty` module is loaded regardless.  I would have expected that to fix and bypass the issue.

Also, it seems `nvim_ui_send` might be the correct API to use instead of `nvim_chan_send`?

BitPusher16:
Can you take a look at detect_from_env()? If you are using ghostty, I would expect that function to return true, and probe() would never need to get called.

PR #540, opened 2026-08-01
Allow clicking outside infoview to switch focus

user202729:
As in the title.

I face the following issue: When the cursor is currently in the infoview window, and I click on another window, it will just be stuck there instead of switching focus to another window (which is the expected behavior).

This pull request fixes that.

The newly added line can be changed to

```
vim.cmd([[exec "normal! \<LeftMouse>"]])
```

which is shorter, but seems less elegant since you need to invoke the Vim interpreter instead of staying entirely inside Lua.

### DIFF
--- a/lua/lean/tui.lua
+++ b/lua/lean/tui.lua
@@ -952,7 +952,7 @@ function BufRenderer:new(obj)
   local function dispatch_mouse(event)
     local pos = vim.fn.getmousepos()
     if pos.winid == 0 or pos.line == 0 or vim.api.nvim_win_get_buf(pos.winid) ~= buf.bufnr then
-      vim.api.nvim_feedkeys(vim.keycode("<LeftMouse>"), "nx", false)
+      vim.api.nvim_feedkeys(vim.keycode '<LeftMouse>', 'nx', false)
       return
     end
     local line_len = #buf:line(pos.line - 1)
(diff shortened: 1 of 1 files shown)


user202729:
Explanation of `n` and `x`:

```
		{mode} is a String, which can contain these character flags:
		'm'	Remap keys.  This is default.  If {mode} is absent,
			keys are remapped.
		'n'	Do not remap keys.
		't'	Handle keys as if typed; otherwise they are handled as
			if coming from a mapping.  This matters for undo,
			opening folds, etc.
		'L'	Lowlevel input.  Other flags are not used.
		'i'	Insert the string instead of appending (see above).
		'x'	Execute commands until typeahead is empty.  This is
			similar to using ":normal!".  You can call feedkeys()
			several times without 'x' and then one time with 'x'
			(possibly with an empty {string}) to execute all the
			typeahead.  Note that when Vim ends in Insert mode it
			will behave as if <Esc> is typed, to avoid getting
			stuck, waiting for a character to be typed before the
			script continues.
			Note that if you manage to call feedkeys() while
			executing commands, thus calling it recursively, then
			all typeahead will be consumed by the last call.
```

The "do not remap key" is important since otherwise you get an infinite loop. The `x` is probably less important, but just in case...?

Note that `vim.keycode` requires Neovim v0.10.0 https://neovim.io/doc/user/news-0.10/

Julian:
Thanks.

